PR #4, opened 2022-03-01
Feature/wketchum hdf5libs integration

wesketchum:
Updated configuration for hdf5libs filelayout configuration in dfmodules.


python/daqconf/apps/dataflow_gen.py
```diff
@@ -30,6 +30,7 @@
 import dunedaq.appfwk.app as app # AddressedCmd,
 import dunedaq.dfmodules.triggerrecordbuilder as trb
 import dunedaq.dfmodules.datawriter as dw
+import dunedaq.hdf5libs.hdf5filelayout as h5fl
 import dunedaq.dfmodules.hdf5datastore as hdf5ds
```
alessandrothea, line 34: Is this import still needed?

wesketchum: I believe so, in order to construct the file layout parameters in the configuration. This is then part of the hdf5datastore configuration.

PR #6, opened 2022-03-01
Migrate code from minidaqapp/nanorc for creating info files describing

eflumerf:
the configuration command line and git tags.

### DIFF
--- a/python/daqconf/core/metadata.py
+++ b/python/daqconf/core/metadata.py
@@ -4,11 +4,6 @@ import sys
 from rich.console import Console
 from os.path import exists, join
 
-CLOCK_SPEED_HZ = 50000000
-
-# Add -h as default help option
-CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
-
 console = Console()
 
 def write_metadata_file(json_dir, generator):
--- a/python/daqconf/core/metadata.py
+++ b/python/daqconf/core/metadata.py
@@ -1,0 +1,35 @@
+import json
+import os
+import sys
+from rich.console import Console
+from os.path import exists, join
+
+CLOCK_SPEED_HZ = 50000000
+
+# Add -h as default help option
+CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
+
+console = Console()
+
+def write_metadata_file(json_dir, generator):
+    console.log("Generating metadata file")
+    with open(join(json_dir, f"{generator}.info"), 'w') as f:
+        daqconf_dir = os.path.dirname(os.path.dirname(os.path.abspath(sys.argv[0])))
+
+        buildinfo_file=join(daqconf_dir, "daqconf_build_info.json")
+        buildinfo = {}
+        #console.log(f"Buildinfo file is {buildinfo_file}")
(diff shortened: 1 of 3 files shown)


bieryAtFnal (approved):
LGTM

plasorak:
I don't understand how this build data should be constructed, but there is a lot of information in the environment:
```
$DAQCONF_DIR        $DAQCONF_INC        $DAQCONF_SHARE      $DAQCONF_VERSION
$DAQCONF_FQ_DIR     $DAQCONF_LIB        $DAQCONFSOURCE_DIR
```

eflumerf:
The build_info is supposed to be populated. The exact directory for daqconf_dir is open for debate, but here's what I get when I installed it in my local UPS products area:
```
{
    "build_info": {
        "build time": "Tue Mar  1 10:46:33 CST 2022",
        "git branch": "eflumerf/AddInfoFiles",
        "git commit author": "Eric Flumerfelt",
        "git commit description": "Migrate code from minidaqapp/nanorc for creating info files describing the configuration command line and git tags.",
        "git commit hash": "d47a20e7eb82e2b9af225b5e25343df3a3ceb612",
        "git commit time": "Tue Mar 1 08:35:05 2022 -0600",
        "hostname for build": "ironvirt7",
        "local repo dir": "/home/eflumerf/Desktop/dune-daq-base/dev-tpset/sourcecode/daqconf",
        "uncommitted changes": "scripts/daqconf_global_gen scripts/daqconf_multiru_gen ",
        "user for build": "eflumerf"
    },
    "command_line": "/home/eflumerf/Desktop/dune-daq-base/products-local/products/daqconf/d47a20e/slf7.x86_64.e19.prof/daqconf/bin/daqconf_global_gen daqconf_multiru --disable-trace --host-thi localhost",
    "daqconf_dir": "/home/eflumerf/Desktop/dune-daq-base/products-local/products/daqconf/d47a20e/slf7.x86_64.e19.prof/daqconf"
}
```

eflumerf:
@plasorak I wanted to avoid depending on the environment variables because they are different when you are running as a UPS product versus when you have the package checked out in `sourcecode` and are running from the local `install` directory.

bieryAtFnal:
It's good that the build information is now populated.

For the daqconf_dir entry, I can imagine this entry meaning "the directory in which the daqconf script was found".  I can also imagine it meaning "the value of the DAQCONF_DIR env var".  But, it seems to be neither of those.  If we don't know its exact meaning, we could say that it is "a general indication of the daqconf area that was used".  That seems a little vague, but if it's the best that we can do....

eflumerf:
OK. I have changed the name to "daqconf_exe_dir" and set it to the directory where the script was actually executing from. I added a comment to the line where the build_info.json is being found to explain the relationship between that directory and the bin directory.

plasorak:
Ok, I supposed I'd prefer if `$DBT_INSTALL_DIR/daqconf/daqconf_build_info.json` was checked for existence and we revert to `$DAQCONF_FQ_DIR/daqconf/daqconf_build_info.json` in case it's not there (I think it would make this script a bit more readable), but your solution should work too.

PR #8, opened 2022-03-01
Fix the FakeDataProd module configuration

eflumerf:
Generated when --use-fake-data-producers is passed to the generation script.


bieryAtFnal (approved):
I have not looked at the code changes, but my tests of the behavior of the system when using FakeDataProd module instances look good.

plasorak:
Ok, so I get data out, I'm not too familiar with the details of the `FakeDataProd` but this seems to work.

PR #14, opened 2022-03-09
Feature/clock speed argument

strilov:


alessandrothea:
The PR description is a little too ephemeral to be able to approve the PR w/o further information...

strilov:
Hi. The PR changes `CLOCK_SPEED_HZ` from a constant defined in `scripts/daqconf_multiru_gen` to an argument passed in at the time of configuration generation.

`CLOCK_SPEED_HZ`is used in several places. It is used to convert the `trigger_rate_hz` argument to the `trigger_interval_ticks` variable. `trigger_interval_ticks` is passed to the HSI modules (fake and real) on conf, start, and resume and is used to control the rate of fake HSIEvents. An issue could arise in a system running at 62.5 MHz with the `FakeHSIEventGenerator` module producing HSIEvents, where the `trigger_interval_ticks` is overridden by the user in nanorc. In such a case the user might expect a 1 Hz rate of fake HSIEvents for `trigger_interval_ticks` of 65000000. However, currently the rate of fake HSIEvents would end up being 1.25 Hz.

`CLOCK_SPEED_HZ` is also used by `DQMProcessor` to convert between ticks and time. It is also used in `tpwriter_gen.py` to convert from time to ticks.

The PR also fixes a bug in hsi_gen.py, where `trigger_interval_ticks` calculated at configuration generation time didn't make into the resulting json configuration files.

alessandrothea (approved):

PR #16, opened 2022-03-10
Use global link number for tpsets links

bieryAtFnal:
This appears to fix the problem with ruemu failing to bind when there
are multiple ruemus running on different hosts


bieryAtFnal:
Both Pierre and Marco have confirmed that this fix addresses the problem.

PR #17, opened 2022-03-11
Bugfix/globalconf opmon cern destinations

strilov:


plasorak:
This seems all correct :) I tried it and it worked.

PR #20, opened 2022-03-24
HSI trigger type passthrough changes (ICEBERG)

MRiganSUSX:
Command-line argument --hsi-trigger-type-passthrough added which affects the trigger decision type assignment in the trigger package.


philiprodrigues:
Looks good to me, and passes my simple tests

PR #21, opened 2022-03-24
Configuration service

plasorak:
This PR makes use of the service described in https://github.com/DUNE-DAQ/pocket/pull/21 and https://github.com/DUNE-DAQ/microservices/pull/1 to access the configuration at NP04.
There are 3 main changes:
 - The addition of the `-u,--upload` flag and `--url` parameter to `daqconf_multiru_gen` to specify if the user wants to upload the configuration to the configuration service listening at `url`. 
 - The addition of `scripts/upload_conf` to upload configurations from a directory structure.
 - The addition of `scripts/download_conf` to retrieve configurations (or list them).

Default URLs will need to be modified.

### DIFF
--- a/python/daqconf/core/upload_conf.py
+++ b/python/daqconf/core/upload_conf.py
@@ -1,0 +1,72 @@
+from appfwk.conf_utils import write_json_files
+import requests
+from pathlib import Path
+import tempfile
+import os
+import json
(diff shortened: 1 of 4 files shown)


plasorak:
This shouldn't go here.

PR #23, opened 2022-03-30
Updated the casual-user docs to reflect the correct steps for v2.10.x

bieryAtFnal:


jcfreeman2 (approved):
Looks good, especially with the updates to account for daq-buildtools best practices (`dbt-env.sh` to set up a work environment in an existing work area, `-c` for a speedier creation of a work area, etc.)

PR #25, opened 2022-04-06
A version of the timing master which works with extra_commands

plasorak:
One needs https://github.com/DUNE-DAQ/appfwk/pull/219 to get this working.


bieryAtFnal:
Target Release?  (I've added this PR to the DUNE-DAQ Tracking project, so attributes like Target Release and Working Group are now available.)

plasorak:
The Target release is v2.11.0, but this is going to look different, based on discussion with Giovanna and Alessandro.

PR #27, opened 2022-04-08
Use CPUPinner module from `readoutmodules`

philiprodrigues:
This PR uses the `CPUPinner` DAQModule introduced by https://github.com/DUNE-DAQ/readoutmodules/pull/4 in config. This is all a temporary solution for the spring 2022 VD coldbox tests. 

We pass a json file describing the pinning config for readout via the `--readout-cpu-pin-file` command line argument. When this argument is passed, a `CPUPinner` module is created in the readout process. The module is created in every readout process, which is not correct for multiple readout processes, but there'll only be one readout process for the VD coldbox, so temporarily, this should be fine.

I've put an example CPU pinning json file in `daqconf/python/daqconf/apps/cpupin.json`, which is probably not the right place, but can move it later

### DIFF
--- a/python/daqconf/apps/cpupin.json
+++ b/python/daqconf/apps/cpupin.json
@@ -1,0 +1,18 @@
+[
+    {
+        "name": "fakeprod-0",
+        "cpu_set": [0]
+    },
+    {
+        "name": "postprocess-0-0",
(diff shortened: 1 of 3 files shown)


roland-sipos (approved):
LGTM

PR #31, opened 2022-04-13
Added fixed=true parameter to hsi network connection as this connecti…

### LINKED ISSUE #30: Fixed connection for global partition
Related to https://github.com/DUNE-DAQ/networkmanager/issues/15
This connection would need to be specified as fixed, as it is a connection to the global partition, and hence shouldn't change depending on which partition one is: https://github.com/DUNE-DAQ/daqconf/blob/develop/scripts/daqconf_multiru_gen#L237

carloschavez:
…on is from a global partition

This change needs to be carefully tested and is related to https://github.com/DUNE-DAQ/networkmanager/issues/15
and the changes in https://github.com/DUNE-DAQ/nanorc/pull/68


eflumerf (approved):
LGTM

PR #33, opened 2022-04-14
Add CPU pinning configuration for np04-srv-030 with data from FELIX

philiprodrigues:
This draft PR is just a placeholder so I can put the committed file somewhere other people can see it.

This file provides a CPU pinning configuration that can be passed to
daqconf_multiru_gen via the --readout-cpu-pin-file option, suitable
for use on np04-srv-030 with the VD coldbox. It's been tested with
emulated data from the FELIX card, and works with up to 7 links.

For each link, there are four threads: ept-0-X-Z, postprocess-0-X,
ind-hits-0-X and consumer-X. The general strategy is to keep all of
the threads on the same NUMA node, with the ept and consumer threads
sharing hyperthread sibling cores (since those two threads have the
lowest CPU usage). Link 3 is an exception: it's split between NUMA
nodes, with the ept and consumer threads on node 0 (even-numbered
cores) and the postprocess and ind-hits threads on node 1.

### DIFF
--- a/python/daqconf/apps/cpupin-np04-srv-030-emudata.json
+++ b/python/daqconf/apps/cpupin-np04-srv-030-emudata.json
@@ -1,0 +1,140 @@
+[
+    {
+        "name": "flx-dma-0",
+        "cpu_set": [0]
+    },
+
+
+
+    {
+        "name": "ept-0-0-0",
+        "cpu_set": [2]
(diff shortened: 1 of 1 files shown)


philiprodrigues:
This is obsolete now as the CPU pinning method has changed

PR #36, opened 2022-04-21
Use `T[PAC]Buffer` for TP/TA/TC buffering in trigger

philiprodrigues:
This PR adds the use of `T[PAC]Buffer` modules from `trigger` in the trigger app, to add TA and TC buffering in addition to the TP buffering. Relies on https://github.com/DUNE-DAQ/trigger/pull/110 .

There are some outstanding issues/questions:

* When the readoutlibs buffer implementation receives a request for some window of data, it needs to see an item past the end of the request window to know that it has got all of the items in the request window. If an item past the end of the window is not seen within a timeout, an empty fragment is returned. This is a problem for low-rate items like TCs. For the moment, I've just increased the timeout for TAs and TCs to 10s (via `retry_count = 1000`), but we need a proper solution in the long term.
* We need some region/element assignment for the TAs and TCs. For now, I've assigned TAs to have region_id = the appropriate one, and element_id = 10,000, to avoid collisions with the TPs. I've assigned TCs (which span the whole partition) to have region_id = 20,000 and element_id = 0. Better ideas are welcome


philiprodrigues:
This PR now also includes changes needed to make the trigger work with the unified communications API

PR #39, opened 2022-04-22
add the thread pinning

plasorak:


philiprodrigues:
I've tested this and see threads pinned to the cores I requested

Issue #22, opened 2022-03-25
Instructions don't work


jcfreeman2:
Right now, if I follow the steps in `InstructionsForCasualUsers.md`, I end up with 
```
-bash: daqconf_multiru_gen: command not found
```
since `dunedaq-v2.8.0`, which users are told to set up with, doesn't have the `daqconf` package. Solutions would be to either (A) refer the reader to the `daq-buildtools` documentation in order to learn how to set up a work area or (B) update the frozen release in the instructions.

jcfreeman2:
For completeness' sake, I'll add that `ConfigurationsForCasualUsers.md` also has the same issue.

PR #41, opened 2022-04-26
Enforce distinct region IDs for each readout app

philiprodrigues:
This PR changes the behaviour of the `--region-id` argument to `daqconf_multiru_gen` so that the region IDs of readout applications are distinct. The old behaviour was:

1. Specify `--region-id` zero times: all readout apps have region ID of zero
2. Specify `--region-id` once: all readout apps have region ID with the given value
3. Specify `--region-id` `n_readout_apps` times: readout apps have the region ID values given by the `--region-id` arguments
4. Specify `--region-id` any other number of times: error

The new behaviour is:
1. Specify `--region-id` zero times: region IDs are numbered sequentially from zero
2. Specify `--region-id` `n_readout_apps` times: readout apps have the region ID values given by the `--region-id` arguments. A check is made to ensure the values are all distinct
3. Specify `--region-id` any other number of times: error

### DIFF
--- a/python/daqconf/apps/trigger_gen.py
+++ b/python/daqconf/apps/trigger_gen.py
@@ -93,15 +93,16 @@ def get_trigger_app(SOFTWARE_TPG_ENABLED: bool = False,
 
         # Make one heartbeatmaker per link
         for ruidx, ru_config in enumerate(RU_CONFIG):
+            region_id = ru_config['region_id']
             for link_idx in range(ru_config["channel_count"]):
-                modules += [DAQModule(name = f'channelfilter_ru{ruidx}_link{link_idx}',
+                modules += [DAQModule(name = f'channelfilter_ru{region_id}_link{link_idx}',
                                           plugin = 'TPChannelFilter',
-                                          connections = {'tpset_sink': Connection(f'heartbeatmaker_ru{ruidx}_link{link_idx}.tpset_source')},
+                                          connections = {'tpset_sink': Connection(f'heartbeatmaker_ru{region_id}_link{link_idx}.tpset_source')},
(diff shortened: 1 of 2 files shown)


bieryAtFnal:
This sounds reasonable to me.  However, I'm curious how it might relate to the setting of other options.  For example, what if I want 2 ReadoutApps, one emulating TPC data and one emulating PDS data.  How would I specify that?  And, could I then specify the same region_id for both of them?
I have similar questions about how we specify certain links to be read out (or emulated).  If I want to specify that links 0, 2, and 4 should be read out, how would I do that?  And, how would I specify the Link IDs for each of them.
I realize that these questions might not be very helpful in getting the practical issue of enforcing unique region IDs resolved in the short term, but I don't know where to raise them, so I'm adding these notes here.

bieryAtFnal:
Given the implications that this Issue might have on configuration management generally, I wonder whether it is something that should be discussed at a CCM meeting (and maybe CCM should be the assigned WG).

glehmannmiotto:
is this correct in general? Once we introduce the PDS, we'll have more than one readout app with the same region ID, no?

dingp:
@philiprodrigues @bieryAtFnal @roland-sipos Should this PR be pushed to 3.1?

bieryAtFnal:
@dingp , I like the idea of having this functionality available in v3.0.0, but there still seem to be some not-yet-understood behaviors in integration tests (e.g. [trigger Issue #22](https://github.com/DUNE-DAQ/trigger/issues/127) and [dfmodules Issue #177](https://github.com/DUNE-DAQ/dfmodules/issues/177), so I'm tempted to push this off to v3.1.0.

eflumerf:
Since there are merge conflicts that seem a little tricky to resolve, I went ahead and made #116 with some of the ideas discussed here.

PR #43, opened 2022-04-27
Small cleanup

jmcarcell:
When working on the previous PR I noticed that this block is repeated (the original is a few lines above), it doesn't do anything because `cardid` and `host_id_dict` are never used again and resetted respectively and the last change to this block was done at the beggining when we were changing to the new configuration generation. I did some quick tests and the configuration generates and works fine


plasorak:
I dont see any problem why this shouldn't be merged.

PR #46, opened 2022-05-11
Feature/hw state check iomanager

strilov:
see issue: https://github.com/DUNE-DAQ/timinglibs/issues/47


bieryAtFnal (approved):
I've run tests of the modified code, and I no longer see error messages at the beginning of runs at ICEBERG that use the TLU/HSI.

PR #47, opened 2022-05-12
Set the 's0' readout window for TimingTriggerCandidateMaker

philiprodrigues:
This has the effect of passing the trigger "before" and "after"
windows from the command line through to the actual TD when the HSI
passthrough option is used, since TTCM uses the s0 readout window for
the TC it creates in HSI passthrough mode.

This fixes https://github.com/DUNE-DAQ/trigger/issues/119


bieryAtFnal (approved):
LGTM.
I confirmed that the two confgen statements listed in the description of the associated Issue in the `trigger` repo generate configs that produce data with different readout windows.

PR #51, opened 2022-05-17
Remove the partition from the configuration 

plasorak:


eflumerf (approved):
I have tested this and it all appears to work

PR #55, opened 2022-05-18
Rename DataLinkHandler timesync name to what dqm is expecting

philiprodrigues:
DQM appears to expect that its timesync connection is named like `timesync_<number>`, but a previous PR from me changed it to avoid collisions with the TP DataLinkHandler. This PR changes back the name of the timesync connection to what DQM is expecting. The TP DataLinkHandler timesync connection keeps the new name, so there are no name collections there.

This needs checking by @jmcarcell - there is one timesync connection per _link_ in readout, but it appears there's one timesync connection per DQM app, so I don't know if the mapping I've got here is right (although I think it's what was there before).


bieryAtFnal:
This seems to help, in the single test that I was able to run.
Unfortunately, when I tried to run a 3ru-by-3df system, I ran into different problems.

bieryAtFnal (approved):
We've seen positive benefits from this change, in that the DQM-related test in readout_type_scan.py works.  Juan will look at possible side-effects when he has time.

PR #59, opened 2022-05-19
Check that topic name does not duplicate a connection UID

### LINKED ISSUE #57: Topic and connection name
Code in here: https://github.com/DUNE-DAQ/networkmanager/blob/5094481e85b2212e3adf2839e3eaed6670fb533e/src/NetworkManager.cpp#L70 imply that the topic and the connetion name need to be different to not get a collision. Is this the expected behaviour? If so we should get `daqconf` to throw an error when this happens.

eflumerf:
Resolves #57


plasorak (approved):
LGTM

PR #62, opened 2022-05-19
Fix timesync connection in DQM

eflumerf:


bieryAtFnal (approved):
With these changes and associated ones in the dqm repo, I confirm that the DQM part of the 3ru_3df integtest in dfmodules now works.

PR #70, opened 2022-05-23
Eflumerf/topo sort for endpoints

eflumerf:
Phil noticed that `make_app_deps` in `conf_utils.py` was no longer functioning correctly, and we were not getting correct start/stop orders as a result.

This PR adds the ability to specify on either end of an endpoint that it should not be used for topological sorting, and for `make_app_deps` to use `System.make_digraph` so that system graph creation is always done in only one place.


philiprodrigues (approved):
The code LGTM, and in testing, this gives a start/stop order that fixes the problem seen in https://github.com/DUNE-DAQ/dfmodules/issues/177

PR #71, opened 2022-05-23
Better tracking of publisher and subscriber connectionids to avoid ad…

eflumerf:
…ding duplicate connectionIDs to the configuration


strilov:
LGTM

PR #76, opened 2022-05-24
K8s

### LINKED ISSUE #56: Add k8s information to boot.json
`daqconf` should generate enough information in boot.json such that `nanorc` can "easily" boot applications in k8s.

plasorak:
Changes in the configuration required for k8s

### DIFF
--- a/python/daqconf/core/conf_utils.py
+++ b/python/daqconf/core/conf_utils.py
@@ -486,7 +486,7 @@ def make_unique_name(base, module_list):
 
 def generate_boot(apps: list, ers_settings=None, info_svc_uri="file://info_${APP_ID}_${APP_PORT}.json",
                   disable_trace=False, use_kafka=False, verbose=False, extra_env_vars=dict(),
-                  image="") -> dict:
+                  image="", external_connections=[], mount_dirs={}) -> dict:
     """Generate the dictionary that will become the boot.json file"""
 
     if ers_settings is None:
@@ -522,13 +522,20 @@ def generate_boot(apps: list, ers_settings=None, info_svc_uri="file://info_${APP
                 "TIMING_SHARE": "getenv",
                 "LD_LIBRARY_PATH": "getenv",
                 "PATH": "getenv",
-                # "READOUT_SHARE": "getenv",
-                "TRACE_FILE": "getenv:/tmp/trace_buffer_${HOSTNAME}_${USER}",
+                "TRACE_FILE": "getenv:/tmp/trace_buffer_$(HOSTNAME)_$(USER)",
+                "CMD_FAC": "rest://localhost:$(APP_PORT)",
+                "INFO_SVC": info_svc_uri,
             },
-            "cmd": ["CMD_FAC=rest://localhost:${APP_PORT}",
-                    "INFO_SVC=" + info_svc_uri,
-                    "cd ${APP_WD}",
-                    "daq_application --name ${APP_NAME} -c ${CMD_FAC} -i ${INFO_SVC}"]
+            "cmd":"daq_application",
+            "args": [
+                "--name",
+                "$(APP_NAME)",
+                "-c",
+                "$(CMD_FAC)",
+                "-i",
+                "$(INFO_SVC)"
+            ],
+            "work_dir": "$(APP_WD)",
         }
     }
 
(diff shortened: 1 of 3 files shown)


python/daqconf/core/conf_utils.py
```diff
@@ -361,7 +361,20 @@ def make_system_connections(the_system, verbose=False):
                     the_system.connections[publisher] += [pubsub_connectionids[connid]]
 
 
+def simplify_connections(app, connections):
```
alessandrothea: Hang on, why the need to "simplify" the connections? Why aren't they created with the right uri in the first place?

python/daqconf/core/conf_utils.py
```diff
@@ -426,8 +439,8 @@ def make_app_command_data(system, app, appkey, verbose=False):
     mod_specs = [ mspec(mod.name, mod.plugin, app_connrefs[mod.name]) for mod in app.modulegraph.modules ]
 
     # Fill in the "standard" command entries in the command_data structure
-
-    command_data['init'] = appfwk.Init(modules=mod_specs, connections=system.connections[appkey])
+    command_data['init'] = appfwk.Init(modules=mod_specs,
```
alessandrothea, line 451: See above. There should be no need to simplify.

alessandrothea (changes requested):

python/daqconf/core/conf_utils.py
plasorak: Resolved

alessandrothea (approved):

python/daqconf/core/conf_utils.py
```diff
@@ -491,36 +510,29 @@ def generate_boot(apps: list, ers_settings=None, info_svc_uri="file://info_${APP
...
-                    "cd ${DBT_AREA_ROOT}",
-                    "source dbt-env.sh",
-                    "dbt-workarea-env",
-                    "cd ${APP_WD}",
-                    "daq_application --name ${APP_NAME} -c ${CMD_FAC} -i ${INFO_SVC}"]
-        },
         "daq_application" : {
-            "comment": "Application profile using  PATH variables (lower start time)",
+            "comment": "Application profile using PATH variables (lower start time)",
+            "image": image,
```
alessandrothea: Is `image` supposed to be defined for k8s configurations as well?

python/daqconf/core/conf_utils.py
```diff
@@ -491,36 +510,29 @@ def generate_boot(apps: list, ers_settings=None, info_svc_uri="file://info_${APP
...
+            "args": [
+                "--name",
+                "{APP_NAME}",
+                "-c",
+                "{CMD_FAC}",
+                "-i",
+                "{INFO_SVC}"
+            ]
         }
     }
```
alessandrothea, line 543: I'm puzzled: control port numbers are supposed to be fixed in kubernetes and staggered when running over ssh. But there is no distinction here?

python/daqconf/core/conf_utils.py
plasorak: `image` is only used in k8s configuration

python/daqconf/core/conf_utils.py
plasorak: Indeed, I don't think there is any reason why it shouldn't be staggered for k8s too.

python/daqconf/core/conf_utils.py
alessandrothea: I meant to say "for ssh configurations" sorry. I assume it's not used by ssh configurations, which makes its present in the `daq_application` profile very confusing.

python/daqconf/core/conf_utils.py
alessandrothea: Because it's not how kubernetes works. With each application encapsulated in a container, identified by a different DNS name, all command (and other service) ports must be the same.

alessandrothea (changes requested):

python/daqconf/core/conf_utils.py
plasorak: Yeah, indeed, and in fact this parameter is completely ignored by k8spm.py https://github.com/DUNE-DAQ/nanorc/blob/plasorak/k8s-np04-cluster/src/nanorc/k8spm.py#L549
So the prefered way to fix this would be to create a new `daq_application_k8s` section in `boot.json` were the k8s information is separated?

plasorak: Actually sorry this is more relevant for the `image` comment. For the commands ports, I don't know how to sort it without differentiating between k8s/non-k8s configurations, which I thought was something we wanted to avoid.

python/daqconf/core/conf_utils.py
plasorak: Ok, this could be fixed by adding a section `daq_application_k8s` in boot.json.

python/daqconf/core/conf_utils.py
alessandrothea: Right, keeping the same configuration is desirable but at a reasonable cost. Clearly we're in a situation where we don't have a way to keep a single consistent configuration w/o injecting confusing/contradictory information.
Regarding `k8spm` ignoring the command port, it is a hack from the prototyping days! Silently overriding configuration parameters is very likely to lead to errors/confusion and must be avoided (for real). 
So I don't see a way not to differentiate between ssh and kubernetes in daqconf, at this stage

plasorak:
Ok, I believe the port and image comments are fixed by the addition of a `--use-k8s` flag on the scripts.

alessandrothea (approved):

PR #78, opened 2022-05-24
Change the connections ignored by the topographical sort so that the …

eflumerf:
…DFO is stopped after the Dataflow.


bieryAtFnal (approved):
These changes helped eliminate warning messages from the DFO about stale TD messages, so they look good to me.

PR #88, opened 2022-06-02
Change polarity of toposort parameter to add_endpoint so that endpoints

eflumerf:
must be explicitly marked as participating in sorting. Add forced_deps
parameter to help resolve ordering conflicts.

Reorder start so that DQM is started after associated apps, while
maintaining trigger start after dataflow.


dingp:
Kurt confirmed in the `integration` channel that:

```
there are no warnings or errors in the logs from the DQM parts of the 3ru_3df and 3ru_1df integtests
when using Eric's changes in daqconf PR #88.  

(still tazipper warning messages, poking around a bit with those now)
```

dingp (approved):

PR #89, opened 2022-06-02
Allow control of the base command port value from commandline

alessandrothea:
The base/default command port is hardcoded to `3333` in `generate_boot`. This leads to conflict when applications from multiple configurations are running on the same host. This PR changes the command port base value into  a`generate_boot` argument and exposes to as a script option

### DIFF
--- a/python/daqconf/core/conf_utils.py
+++ b/python/daqconf/core/conf_utils.py
@@ -478,7 +478,7 @@ def make_unique_name(base, module_list):
 
     return f"{base}_{suffix}"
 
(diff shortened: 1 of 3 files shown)


dingp (approved):
LGTM

dingp:
Verified that the new command line option works, and as well as the default value with:

`daqconf_multiru_gen --host-ru localhost --latency-buffer-size 200000 -d ./frames.bin -o . -s 10 mdapp_5proc`
`daqconf_multiru_gen --host-ru localhost --latency-buffer-size 200000 -d ./frames.bin -o . -s 10 mdapp_5proc --base-command-port 3334`

And `nanorc mdapp_5proc boot ${USER}_test init conf start 111 wait 60 stop scrap terminate` succeeded for both configs.

PR #94, opened 2022-06-07
Offset link numbers for real datahandlers only

ShyamB97:


bieryAtFnal (approved):
These changes look good to me, in terms of not affecting existing functionality that is tested in the dfmodules integration tests.

PR #97, opened 2022-06-08
offset link numbers for firmware-tps only

ShyamB97:
Link numbers were being offset when --enable-software-tpg was flagged, likely causing the large scale test to fail for n > 5 linkhandlers.


ShyamB97:
large scale tests passed using this branch, @bieryAtFnal is it worth running integration tests before merging?

bieryAtFnal:
sorry, I've been away from the keyboard for a while.  I'll run some integtests now.

bieryAtFnal (approved):
My tests of these code changes, using the integration tests in dfmodules/integtest, do not show any new problems.
As such, these changes look OK to me.

PR #99, opened 2022-06-09
Adding boolean option to trigger_gen.py for switching trigger types.

CharlieBatchelor:
Required boolean options to switch trigger type when configuring the trigger algorithms.


dingp:
Converted the PR to draft since we do not want this to be merged to `develop` before we freeze `dunedaq-v3.0.0` release.

PR #106, opened 2022-06-15
Restore Correct DFO host

mroda88:
Based on what we discussed at the CoreSW meeting on 15 June 2022, we Reverts DUNE-DAQ/daqconf#104 .


ron003 (approved):
looks good -- I'll merge into develop

PR #107, opened 2022-06-15
sync with develop to bring in the fix for DFO host

mroda88:


ron003 (approved):
simple diff -- merge into new patch branch

PR #110, opened 2022-06-16
Dynamic queue size between TRB and DataWriter to handle long-window-readout

bieryAtFnal:
In tests of long-window-readout, Wes and I have noticed there can be complaints in the dataflow processes about failing to push an element onto a queue within the specified timeout.  The queue in question is the one between the TriggerRecordBuilder and the DataWriter, and its default size is 10.

This problem happens when there is a large burst of TriggerRecords associated with a single long-window-readout trigger.  For example, with a 1 second readout window and a max window size per TR of 4 msec, there will be 250 TriggerRecords associated with each trigger, and those TRs can appear in the TRB very quickly.

The code changes in this PR make the size of the TRB-DW queue dynamic, based on the maximum number of TRs expected per trigger (plus a little safety factor).

### DIFF
--- a/python/daqconf/apps/dataflow_gen.py
+++ b/python/daqconf/apps/dataflow_gen.py
@@ -45,6 +45,7 @@ def get_dataflow_app(HOSTIDX=0,
                      TPC_REGION_NAME_PREFIX="APA",
                      MAX_FILE_SIZE=4*1024*1024*1024,
                      MAX_TRIGGER_RECORD_WINDOW=0,
+                     MAX_EXPECTED_TR_SEQUENCES=1,
                      HOST="localhost",
                      HAS_DQM=False,
                      DEBUG=False):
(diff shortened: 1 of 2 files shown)


bieryAtFnal:
Here is a sample confgen command for testing this...
```
daqconf_multiru_gen -d $PWD/frames.bin -o . -s 10 -n 2 -b 99000000 -a 1000000 -t 0.2 --latency-buffer-size 600000 --host-ru localhost --host-df localhost --region-id 0 --max-trigger-record-window 100000 -c 1 lwr_stresstest
```

Also, I would typically use a non-trivial resume-wait time (10-20 seconds) when I would start a run for this testing to allow the latency buffers to fill with emulated data.

mroda88 (approved):
I gave it a try. I generated the configuration suggested by Kurt with both v5.3.1 and with this branch, that order. 
Screenshot from 2022-06-16 16-05-42

As you can see this branch prevents the queue to be saturated and the run seems indeed more stable, according to other metrics.

PR #114, opened 2022-06-20
K8s config

plasorak:
With this PR and the corresponding in nanorc, daqconf is responsible for generating
 - boot order
 - node affinity
 - pod (anti) affinity
Also the host dictionary now is defined as "app": "host_name" rather than "host_app": "host_name".

### DIFF
--- a/python/daqconf/core/app.py
+++ b/python/daqconf/core/app.py
@@ -20,7 +20,7 @@ class ModuleGraph:
 
     def combine_queues(self, queues : [Queue]):
         output_queues = []
-        
+
         for q in queues:
             match = False
             for oq in output_queues:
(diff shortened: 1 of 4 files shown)


plasorak:
Ok, so integration test do work on this (after yesterday's fix on the DQM)
Basically, this PR:
 - Adds some members to the `App` class that are specific to K8s. https://github.com/DUNE-DAQ/daqconf/blob/plasorak/k8s-config/python/daqconf/core/app.py#L229
 - Modifies `boot.json` such that there is much less hardcoding in nanorc.

More specifically the following k8s configuration items are now available to configuration:
 - The node selection can now be done in a smarter way, and enable use of the labels in k8s. Right now, when the host is specified, it sets `kubernetes.io/hostname`, but that can be expended. One can toggle how strict this requirement is with the `strict` flag. If the `strict` flag is `False`, the node is "preferred," so k8s will try to assign it to the node, and if it cannot do that, start it somewhere else. With `strict == True`, the app won't boot if the node isn't available (for example it has disk pressure...). By default, this flag isn't `True`, and needs to be set manually in the `daqconf_multiru_gen` script (I've done it for the readout and timing, maybe not in the nicest way).
 - Pod affinity and anti affinity, i.e. whether the apps need to run on the same or different nodes (again you can specify strict or not). For example [this](https://github.com/DUNE-DAQ/daqconf/blob/plasorak/k8s-config/scripts/daqconf_multiru_gen#L513) sets the `dqm_df` apps to preferrably run with the `dataflow` apps.
 - `resources`: `daqconf` is now able to specify what resources the application is using, [one example from FELIX](https://github.com/DUNE-DAQ/daqconf/blob/plasorak/k8s-config/scripts/daqconf_multiru_gen#L409).
 - PVCs: For that, we now have a local-path-provisioner at NP04 (that runs in the namespace `raw-data-storage`). An example is [here](https://github.com/DUNE-DAQ/daqconf/blob/plasorak/k8s-config/scripts/daqconf_multiru_gen#L264) for how to set it. Basically, the `output-dir` is now an uri, which has the form `storage-class-name:/directory`.

Another thing that I simplified is the `host` dictionary (so from `host_ruemu0` to `ruemu0`). That removes some dirty code in `nanorc`.

PR #116, opened 2022-06-22
Eflumerf/region id and topics

eflumerf:
Replaces https://github.com/DUNE-DAQ/daqconf/pull/41 (mostly)

Change TPSet connections to use "TPSets" as their topic. Make TPWriter use the "TPSets" topic as the input.

Update --region-id to assign incremental region IDs to RUs if not
specified on command line.

Note that multi-frontend systems are not currently supported by
daqconf_multiru_gen, and when they are, the region-id logic may need
updates.

### DIFF
--- a/python/daqconf/apps/readout_gen.py
+++ b/python/daqconf/apps/readout_gen.py
@@ -145,7 +145,7 @@ def get_readout_app(RU_CONFIG=[],
                                           emulator_mode = EMULATOR_MODE,
                                           error_counter_threshold=100,
                                           error_reset_freq=10000,
-                                          tpset_topic=RU_CONFIG[RUIDX]["tpset_topics"][idx]
+                                          tpset_topic="TPSets"
                                       ),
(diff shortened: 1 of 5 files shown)


bieryAtFnal:
These changes look good to me.  I've run automated (dfmodules) integration tests and haven't seen any issues.  A visual inspection of the TPSet connections in the generated configuration with these changes shows reasonable connection names, subscriptions, etc.

Test runs with the existing daqconf code and this new daqconf code produce similar raw data files.  The TPStream file is larger with these changes, which is good.  (The existing code seems to produce configurations with typos, so that problem probably affected the runs with the existing code.)

```
daqconf_multiru_gen --host-ru localhost --host-ru localhost --host-df localhost --host-df localhost -d $PWD/frames.bin -o . -s 10 -n 2 --enable-software-tpg --enable-tpset-writing mdapp_2RU_2DF_swtpg_tpsw_new
```

bieryAtFnal (approved):
LGTM, and Phil is OK with us proceeding with these changes.

PR #118, opened 2022-06-23
Tweak topographical sort (and add forced dependencies)

### LINKED ISSUE #115: Revise stop order
From Core SW WG Meeting June 22, 2022:

Stop order now : [
        "ruemu0",
        "trigger",
        "dfo",
        "dataflow0",
        "hsi"
    ]

Desired stop order: DFO, dataflow, readout, HSI, Trigger

    try to capture the rationale afterward

    pause to Trigger stops the generation of triggers

    DFO first allows all triggers in flight to be completely read out, built, and written

This relies on https://github.com/DUNE-DAQ/dfmodules/issues/198 so that once the DFO is stopped we know that all Trigger Records for the current run have been written.

eflumerf:
To achieve desired stop order where DFO stops first, followed by dataflow, readout,
HSI, and finally trigger.

The rationale for this order is that trigger should have stopped issuing
triggers at pause, but needs to be around to respond to data requests
from other parts of the system. DFO will wait until all TRs are complete
before allowing the rest of the system to stop.

Resolves #115. Depends on https://github.com/DUNE-DAQ/dfmodules/pull/201


bieryAtFnal (approved):
Several of us have run successful tests of these changes.

PR #121, opened 2022-07-01
Reduce buffer timeouts and don't warn on TC data request timeouts

philiprodrigues:
To implement the behaviour requested in https://github.com/DUNE-DAQ/trigger/issues/134 , this PR reduces the timeouts on the TP, TA and TC buffers.

Requires https://github.com/DUNE-DAQ/readoutlibs/pull/40 to be merged first


bieryAtFnal (approved):
LGTM

PR #123, opened 2022-07-01
Fixed the calculation of the size of the TRB-DataWriter queue

bieryAtFnal:
based on the number of TR sequences...


eflumerf (approved):
LGTM

PR #124, opened 2022-07-07
Data request timeout parameter modified to use a time interval instead of a retry count

bieryAtFnal:
These changes are a reaction to changes in the readoutlibs repo, and the merges to develop will need to happen at the same time (after review, etc).


jcfreeman2 (approved):
LGTM.

PR #127, opened 2022-07-12
Add PD2HD channel map

jmcarcell:
Maybe for the future we can use the same names in DQM and for tpg (readout and trigger)


philiprodrigues (approved):
LGTM

PR #128, opened 2022-07-13
Add Kafka entry for opmon stream to replace the influx stream

mroda88:
This PR requires the dedicated microservice to be in place, not just in the code, but in the cluster as well. 
I think we should wait until that ready before merging. that's why this is a draft PR.


mroda88:
This should solve https://github.com/DUNE-DAQ/kafkaopmon/issues/3

mroda88:
Before merging this we should first merge https://github.com/DUNE-DAQ/kafkaopmon/pull/5 
Yet the microservice that allows this is up and running, so the PR is ready for review.

PR #134, opened 2022-07-15
Adding parameters for trigger-MLT: timeouts and readout limits

MRiganSUSX:
Adding 3 configurable parameters:
-> mlt-buffer-timeout: buffer period to wait for new overlapping TCs to merge into single TD
-> mlt-td-out-of-timetout: flag, to either keep or drop TCs that arrived after TD was sent but overlap with the readout window
-> mlt-td-readout-limit: a maximum length of the readout window for a single TD


philiprodrigues:
Converted this to draft because it's aimed at 3.2

PR #139, opened 2022-07-19
Plasorak/timing gen bugfix

bieryAtFnal:
These changes are needed to get the generation of the configuration for the global timing partition to work.


plasorak (approved):
I think this is all fine. The node selection stuff is anyway going to change.

PR #140, opened 2022-07-21
Change the default argument in generate_boot_common

philiprodrigues:
Default argument for `daq_app_exec_name` should be `"daq_application_ssh"` not `"daq_application"`


plasorak (approved):
Yep, all good and a bit simpler for everybody.

PR #142, opened 2022-07-21
Strict host assignement for k8s

plasorak:
This PR makes the the K8s system SSH-like, by strictly assigning the pod to the node specified, except if that node isn't specified (i.e. the default: localhost).
There are a some stuff from https://github.com/DUNE-DAQ/daqconf/pull/141 which should be merged before


carloschavez (approved):
looks good to me

PR #143, opened 2022-07-21
Mount raw data storage on the pod

plasorak:
This PR scraps the PVC/PV and rancher utility that were used for raw data storage, and directly mounts the output dir on the dataflow and tpwriter.
There are stuff from https://github.com/DUNE-DAQ/daqconf/pull/141 which are needed to use it.


carloschavez (approved):
We tested these changes with a few configurations. it worked as expected

PR #146, opened 2022-07-21
Make FakeDataProd use WIB2 frame size constants unless clock is set to 50 MHz (WIB1 clock)

### LINKED ISSUE #144: fake data prod generation sets wrong time ticks if clock is 62.5 MHz
If the clock is 62.5 MHz, the fake data producers configuration should set 32 at the time_ticks_diff, as a wib emulator

eflumerf:
Resolves #144


glehmannmiotto (approved):
Thanks Eric, looks good.

PR #148, opened 2022-07-21
Initial casual user instruction update for v3.1.0

bieryAtFnal:
The v3.1.0 diagram that I found had text that was obsolete (talked about new features that we have deferred until later), and it had the same blocks in the diagram as the v3.0.0 version, so I didn't update the diagram.
And, of course, I'm guessing what the `setup_dbt` and `dbt-create` arguments should be, based on what we had in the past.


jcfreeman2 (approved):
LGTM.

PR #153, opened 2022-07-27
Make HSI input and use of channel filter optional

philiprodrigues:
The motivation for this is so we can use the application returned by
`get_trigger_app()` in trigger tests outside of the main system. In
particular, for tests where we replay (FD MC) TP data from file, eg
for CPU usage/scaling tests.

It might be nice to just leave the HSI input there and not connect it
to anything, so then there'd be no HSIEvents into the MLT and things
would work as we want. But the daqconf setup complains when there are
unconnected endpoints. (I tried adding a HSI app to the config to get
rid of the daqconf errors, but then I get daqconf errors about there
being nothing connected to HSI's TimeSync input, because there are no
readout apps).

The reason to remove the channel filter is because we don't have a
channel map for FD MC, so running FD MC through the system results in
crashes in the channel filter as it tries to look up channels that
don't exist in the ProtoDUNE/VD coldbox channel map.

### DIFF
--- a/python/daqconf/apps/trigger_gen.py
+++ b/python/daqconf/apps/trigger_gen.py
@@ -73,12 +73,14 @@ def get_trigger_app(SOFTWARE_TPG_ENABLED: bool = False,
                     CANDIDATE_CONFIG: int = dict(prescale=10),
 
                     SYSTEM_TYPE = 'wib',
+                    USE_HSI_INPUT = True,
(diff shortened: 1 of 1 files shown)


CharlieBatchelor (approved):
Tested this branch in conjunction with `replay_tp_to_chain` of trigger for the new fake TP chain app. Everything seems to work as expected.

PR #155, opened 2022-07-28
remove globa timing part generators

strilov:


plasorak (approved):
This is a trivial change (all the files are moved to timinglibs)

PR #162, opened 2022-08-12
added patches to readout_gen from VD coldbox run

ShyamB97:
Adds patches to readout_gen.py to get firmware TP working which were not included in dunedaq-v3.1.0


eflumerf:
Replaced by #165.

PR #163, opened 2022-08-23
set FRONTEND_TYPE in readout such that right modules are loaded from …

glehmannmiotto:
…connection instance names.
This is a temporary fix, before we change to the connectivity service and the readout won't use the connection instance names anymore to load the right plugins.


roland-sipos (approved):
LGTM and actually needed for a succesful run.

PR #166, opened 2022-08-24
Update daqconf for wib2 file format

adam-abed-abud:
Removing the raise of exception when using the wib2 frame. At the moment it cannot be used for the wib2 format.


jmcarcell:
I think it would be better to keep the check and check also for wib2 frames

adam-abed-abud:
Thanks Juan! I added the exception in case the swtpg is enabled when the format is not wib or wib2

eflumerf:
It appears to me that this change is invalidated by the SourceID changes to `daqconf`. @adam-abed-abud, could you check and close this PR if that is the case?

adam-abed-abud:
Hi Eric! Yes this PR is not needed anymore because of the recent `sourceID` changes. I will close the PR.

PR #171, opened 2022-08-31
Improve generation of DFO config by passing in DF app configs (to get

eflumerf:
tokens per-app and correct TRBuilder SourceIDs)


bieryAtFnal:
Question: how does this affect the changes that I'm making to provide a TRBuilder source_id value to the TPStreamWriter?
Would I continue to use lines like the following?
```
    if enable_tpset_writing:
        tpw_name=f'tpwriter'
        **dfidx = sourceid_broker.get_next_source_id("TRBuilder")
        sourceid_broker.register_source_id("TRBuilder", dfidx, None)**
        the_system.apps[tpw_name] = get_tpwriter_app(
            OUTPUT_PATH = tpset_output_path,
            OPERATIONAL_ENVIRONMENT = op_env,
            MAX_FILE_SIZE = tpset_output_file_size,
            DATA_RATE_SLOWDOWN_FACTOR = data_rate_slowdown_factor,
            CLOCK_SPEED_HZ = clock_speed_hz,
            HARDWARE_MAP_FILE=hardware_map_file,
            **SOURCE_IDX=dfidx,**
            HOST=host_tpw,
            DEBUG=debug)
```
Would I still need to move the tpwriter_gen section in daqconf_multiru_gen after the DF section?
Thanks

bieryAtFnal:
This change doesn't seem to work with more than one TRB.
Maybe I'm doing something wrong, but, for example, the 3ru_3df integtest can't generate configs.

eflumerf:
OK, I fixed the problem. The location of the tpwriter block in the config should no longer matter.

bieryAtFnal:
Comparing the generated configurations before and after these changes, I see the following differences in the dfo_conf.json:
```
9,10c9,10
<                             "busy": 20,
<                             "free": 10
---
>                             "busy": 10,
>                             "free": 5
16,17c16,17
<                             "busy": 20,
<                             "free": 10
---
>                             "busy": 10,
>                             "free": 5
```

is this expected? 

This is with two DF apps, SWTPG enabled, and TPset-writing enabled.

eflumerf:
I think the old config was wrong, since each DF app was only configured with 10 tokens. (Specifically the old line 238 in daqconf_multiru_gen) Also, now if you configure one app with 10 tokens and the other with 20, you'll see the difference reflected in dfo_conf.json.

bieryAtFnal (approved):
OK, LGTM.

PR #175, opened 2022-09-07
move boot away from scripts

plasorak:
The changes in this PR generate bootable apps, but I haven't tried with K8s yet.


eflumerf:
Does this still work without specifying a config file? e.g. `daqconf_multiru_gen default`

plasorak:
Probably not, let me have a look

plasorak:
Ok, this is ready now, I'll move on to `listrev` et al.

eflumerf (approved):
Probably my only remaining concern is the change from "daqconf" to "boot", as that will invalidate existing configurations, but I think this looks good otherwise.

plasorak:
Ah good point. This indeed is going to make old json not working. I'll revert this back as I don't think it's a good idea to break everything now. We can keep that for next release (I think ultimately we should have a "common" and a "boot" section).

PR #177, opened 2022-09-08
Fix ND_LAr detector name hack

eflumerf:


eflumerf:
See https://github.com/DUNE-DAQ/detdataformats/blob/9f79db8ddfc8709b501762b10248402c68d0a575/include/detdataformats/detail/DetID.hxx#L86 for correct string

PR #180, opened 2022-09-08
Remove unneeded -t and -f options, add --enable-dqm

eflumerf:
(use_felix should be set in config as it's a major parameter, trigger_rate_hz should be in config or at start_run), add --enable-dqm as it will be needed for testing


plasorak:
Was there a particular reason for letting these as CLI parameters in the first place?

eflumerf:
It was mostly me trying to anticipate options that might be needed without creating a whole new config file. The logic I'm using for adding --enable-dqm is that we will often want to run tests with and without DQM processing, and we don't necessarily want to have different configurations otherwise.

plasorak (approved):

PR #181, opened 2022-09-08
Throw an exception when config file is passed but doesn't exist

eflumerf:


jcfreeman2:
Tested it both with a bogus config filename and a real config filename, and the script did what you'd except. Looks good.

PR #182, opened 2022-09-12
Sbhuller/fwtpg

ShyamB97:
Changes to readout_gen.py to get readout of firmware TPs functional


eflumerf:
Could you also take a look at `https://github.com/DUNE-DAQ/daqconf/blob/develop/python/daqconf/core/sourceid.py`? I think you may want to register the FW TP sourceIDs there (in `register_readout_source_ids` and update the logic in `generate_trigger_source_ids`). Since SourceIDs are just numbers, I don't think that there would be an issue with always registering FW TPG source IDs, even if the configuration ends up not using them.

NB: `register_source_id` allows you to associate an object with the SourceID in the SourceIDBroker. You can see how this is used in `generate_trigger_source_ids`, and it may be a good idea to mark the FW TPG SourceIDs in a similar fashion, so you can retrieve them later in `readout_gen.py`

ShyamB97:
@eflumerf ok, I have tagged the fw source IDs in the broker, I don't think there are any changes to make in the `generate_trigger_source_ids` no?

eflumerf:
I would think that [this section](https://github.com/DUNE-DAQ/daqconf/blob/1f91f5e8a5747a7775643fd8fb887b4ea7366b1b/python/daqconf/core/sourceid.py#L89) would have to change to mirror the new Source IDs for the FW TPG links, so that the trigger knows which trigger source ids correspond to the FW TPG links

ShyamB97:
@eflumerf ok I the changes are made. I ran some quick tests and fwtp readout seems to work

python/daqconf/apps/readout_gen.py
```diff
@@ -81,15 +81,20 @@ def get_readout_app(DRO_CONFIG=None,
...
-            if not link.dro_slr in link_to_tp_sid_map.keys():
-                link_to_tp_sid_map[link.dro_slr] = SOURCEID_BROKER.get_next_source_id("Trigger")
-                SOURCEID_BROKER.register_source_id("Trigger", link_to_tp_sid_map[link.dro_slr], None)
-                slr_link[link.dro_slr] = link.dro_source_id
+    if FIRMWARE_TPG_ENABLED:
+        for fwsid in SOURCEID_BROKER.get_all_source_ids("FW_TPG"):
+            slr = SOURCEID_BROKER.sourceid_map["FW_TPG"][fwsid]
+            link_to_tp_sid_map[slr] = fwsid
+            slr_link[slr] = link_to_tp_sid_map[slr]
+            SOURCEID_BROKER.register_source_id("Trigger", link_to_tp_sid_map[slr], None)
```
eflumerf: You have to be careful here to call `get_next_source_id`, to ensure you're not trying to re-use a SourceID that was registered elsewhere.

eflumerf:
I don't think these changes will work with multiple hosts/cards. You may need to add more information to the registration object with host/card/slr to uniquely identify each FWTP link.

python/daqconf/apps/readout_gen.py
ShyamB97: so the problem I encountered with `get_next_source_id` is that is starts checking source IDs at zero, but the firmware TP source IDs start at a non-zero value. so when I use `get_next_source_id` it will just return zero. Is there a way to include some offset to `get_next_source_id` so it doesn't just start at zero?

ShyamB97:
> I don't think these changes will work with multiple hosts/cards. You may need to add more information to the registration object with host/card/slr to uniquely identify each FWTP link.

are you referring to this part of `sourcid.py`? In that case should I just make a new info class which contains the relevant information?

eflumerf:
I've put a few of my ideas in #190, if you'd like to take a look and possibly test

eflumerf (approved):
LGTM

PR #187, opened 2022-09-13
Add printout of defaults from schema to daqconf_multiru_gen -h.

eflumerf:
I'm creating this as a draft PR, since there are 2 questions I can think of right off the bat...

1. Is this helpful?
2. Should this be implemented lower down (i.e. config_file.py) and/or better so that it can be used elsewhere?

Unfortunately, click does its own reformatting to the provided option help text, so giving the whole thing results in very ugly output. Moo's `help(<obj>)` as seen [here](https://brettviren.github.io/moo/otypes.html#org9793c0c) is also hard to parse, which is why I manually walk the config structure.

### DIFF
--- a/python/daqconf/core/config_file.py
+++ b/python/daqconf/core/config_file.py
@@ -120,3 +120,18 @@ def parse_config_file(filename, configurer_conf):
             # return parse_ini(filename, configurer_conf), filename
 
     raise RuntimeError(f'Configuration {filename} doesn\'t exist')
+
+def helptree(ost, prefix=''):
+    if 'doc' in ost.keys():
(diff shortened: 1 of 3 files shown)


plasorak (commented):
Thanks, looks good.

plasorak:
Actually, on second thought, indeed it's be good if this was implemented lower down, so that we can do:
```
help = helptext(confgen.daqconf_multiru_gen())+helptext(confgen.dataflowapp())
```
Also, could there be the type of the parameter (I think that's `ost["schema"]`)?

eflumerf:
I looked, but since the types of each parameter (`item` in the schema) are locally-defined, I'm not sure if it actually adds useful information.

PR #189, opened 2022-09-14
less boiler plate

plasorak:


plasorak:
I managed to make the help "generic" so that one doesn't need to explicitely feed `dataflowapp`. This reduces the boiler plate code to a single CLI decorator.

eflumerf (approved):
LGTM. And you solved the weird-wrapping-in-click-help issue!

PR #191, opened 2022-09-14
Use appropriate TP SID for datahandler TPSet stream

eflumerf:


bieryAtFnal (approved):
In a discussion with Eric and Roland earlier today, I believe that we agreed to move forward with these changes, so I will approve them and then merge them.

PR #193, opened 2022-09-15
First configuration using DPDK

jmcarcell:
This is the first iteration of a configuration that adds a frames sender and frames receiver over ethernet using dpdk. 

Can't be merged until dpdklibs is included in the builds


jmcarcell:
Looks like nothing changed when not using the new parameters

PR #194, opened 2022-09-16
add daqconf_DAQSHARE definition into cmake.in, so that other pkgs can…

dingp:
… use daqconf when doing codegen

This is causing `listrev`, which recently added `daq_codegen( confgen.jsonnet DEP_PKGS daqconf TEMPLATES Structs.hpp.j2)`, to fail the nightly build with:

1. Targets file not found (which is reasonable since there's no cmake targets for daqconf);
2. daqconf_DAQSHARE is not defined.


dingp:
Verified locally `listrev` can be built successfully by spack using `daqconf` from spack installation.

PR #195, opened 2022-09-16
bringing k8s configuration uptodate

carloschavez:
Bringing daqconf configuration in line with updates for K8s running


eflumerf (approved):
LGTM

PR #199, opened 2022-09-19
Jbrooke/dtpctrl

ShyamB97:
Add dtpcontroller module to readout_gen to control fwtpg through the daq_app


ShyamB97:
also contains bugfixes from this weeks coldbox run to do with the tp_out linkhandlers

eflumerf (approved):
LGTM. Confgen works correctly and produces reasonable output.

PR #202, opened 2022-09-21
Make HardwareMap and multiple output path work with K8s

plasorak:
... And provide a json example


eflumerf (approved):
As discussed at v3.2.0 release meeting, this is probably how we'll do this for v3.2.0, maybe something better for configuring the HWM service for v3.3.0.

PR #203, opened 2022-09-21
add the RunTimeEnv script

plasorak:


python/daqconf/core/conf_utils.py
```diff
@@ -777,3 +782,45 @@ def write_json_files(app_command_datas, system_command_datas, json_dir, verbose=
...
+    from os import getenv
+    version = getenv("DUNE_DAQ_BASE_RELEASE")
+    if not version:
+        raise RuntimeError('Utils: dunedaq version not in the variable env DUNE_DAQ_BASE_RELEASE! Exit nanorc and\nexport DUNE_DAQ_BASE_RELEASE=dunedaq-vX.XX.XX\n')
+    return version
+
+def nightly_or_release(version):
+    from re import match
+    if   match( 'N[0-9]{2}-[0-9]{2}-[0-9]{2}', version): return 'nightly'
+    elif match('NT[0-9]{2}-[0-9]{2}-[0-9]{2}', version): return 'nightly'
```
eflumerf: I think the letter is arbitrary...maybe `N[A-Z]?[0-9]` would be better

eflumerf: Also candidate releases have a different naming pattern.

plasorak: Done; for the candidate release, provided they have the form `dunedaq-v3.1.1-rc1` would be supported (`.*` at the end, i.e. any number of arbitrary characters). But I'll double check it indeed works.

plasorak: Ok, I've done this even better now.

eflumerf (approved):
LGTM, at least it doesn't interfere with non-k8s configurations

plasorak:
Well, if you run on a release, even with SSH, it will use the RTE

PR #214, opened 2022-09-27
Addition of the option to ignore specific TC types in trigger-MLT

MRiganSUSX:
option to pass a vector of TC types to be ignored in the trigger-MLT
passed via config json
default empty


python/daqconf/apps/trigger_gen.py
```diff
@@ -90,6 +90,7 @@ def get_trigger_app(CLOCK_SPEED_HZ: int = 50_000_000,
                     MLT_BUFFER_TIMEOUT: int = 100,
                     MLT_SEND_TIMED_OUT_TDS: bool = False,
                     MLT_MAX_TD_LENGTH_MS: int = 1000,
+                    MLT_IGNORE_TC: dict = {},
```
philiprodrigues: Is this really a dict? Looks more like it would be a list

philiprodrigues (changes requested):

python/daqconf/apps/trigger_gen.py
MRiganSUSX: Yes, it indeed is a list. Wasn't sure if we want to use list (as it is not used throughout the document) and the dictionary worked the same, but changed it to list now (in next commit). Reran the tests as well and works fine.

MRiganSUSX:
This was already merged into the patch branch in #262

PR #220, opened 2022-09-29
Fix the frontend type not being passed to DQM

jmcarcell:
During the SourceID changes the logic for getting the frontend type was moved to the readout config generation but DQM also needs it. I've assumed the frontend type is the same for every SourceID which I think it's fine for this release but we'll have to change it in the future


eflumerf:
I'm wondering if it would be better to leave the logic at the app_gen level...Of course, we could make a function in `conf_utils.py` to do the conversion.

eflumerf (approved):
LGTM

PR #221, opened 2022-09-29
Fix app names

plasorak:
Fix DQM name
Make throw an error if this happen again


jmcarcell:
Shouldn't it be against the prep branch?

plasorak:
it should yes, thanks

eflumerf (approved):
LGTM

PR #222, opened 2022-09-29
Update app name for fragment_producers check

eflumerf:


plasorak:
The trb module of dqm is `trb_dqm` not just `trb`

plasorak:
Ah NVM, looks like you have it correct

plasorak (approved):
LGTM

PR #224, opened 2022-10-03
Backed out the passing of the enable_firmware_tpg flag to tpwriter_app…

bieryAtFnal:
…since the TPStreamWriter should only write 'SW' TPs.

This change is part of a more thorough fix for dfmodules Issue 231.  It is for consideration **after** v3.2.0.  It is coupled with PR 237 in the dfmodules repo.


eflumerf (approved):
builds, minimal_system_quick_test and tpstream_writing_test both pass all cases.

PR #228, opened 2022-10-05
Sbhuller/dunedaq v3.2.0 rc patch

ShyamB97:
Saw some issues when generating configurations with the dpdk features added to readout_gen during CB tests


eflumerf (approved):
LGTM

PR #229, opened 2022-10-05
Fix a configuration bug in DQM

jmcarcell:
that wasn't getting the right number of ticks for different frontend types


eflumerf (approved):
LGTM

PR #230, opened 2022-10-06
Fix location of FakeCardReader in readout_gen's logic.

eflumerf:
A recent PR made the FakeCardReader insertion logic only happen when `USE_FAKE_DATA_PRODUCERS` was true, but that is the configuration to enable `FakeDataProd` instead of `DataLinkHandler`. When reading a file, (e.g. `FLX_INPUT` and `ENABLE_DPDK_READER` are both false), `FakeDataProd` needs to be enabled.


bieryAtFnal (approved):
LGTM.  minimal_system_quick_test and other integtests now work.

ShyamB97 (approved):
thanks for sorting this out, confgen with the real card + fwtps looks fine.
LGTM

PR #236, opened 2022-10-13
Jhancock/234 configurable rte script usage

### LINKED ISSUE #234: Configurable RTE script usage
Right now, https://github.com/DUNE-DAQ/daqconf/blob/develop/python/daqconf/core/conf_utils.py#L659 assumes that you don't want to use a RTE (run time environment) script setup if you are running with a development area. This is usually the case, but it would be nice if that was configurable in the boot section of the json, so that it can easily be toggled for run control development for example.
So we need a switch, which overrides the `release_or_dev()`check on that line.

JonathanHancock0:
I have added a new possible field in the boot section of the config JSON: "always_RTE_script". If this is false (it defaults to this if unspecified), then RTE scripts will be used if you are NOT in a dev area. If it is true, then a RTE script will be used regardless of any areas.

### DIFF
--- a/python/daqconf/core/conf_utils.py
+++ b/python/daqconf/core/conf_utils.py
@@ -656,8 +656,18 @@ def generate_boot(
     if conf.disable_trace:
         del boot["exec"][daq_app_exec_name]["env"]["TRACE_FILE"]
 
-    if (release_or_dev() == 'rel') or (conf.always_RTE_script):
-        boot['rte_script'] = get_rte_script()
+    match conf.RTE_script_settings:
+        case 0:
+            if (release_or_dev() == 'rel'):
+                boot['rte_script'] = get_rte_script()
+
+        case 1:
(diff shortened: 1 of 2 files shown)


plasorak:
Actually, I think what we want here is a bit more complicated. Could we have an integer (or an enum) that holds the behaviour as:
 - 0: Figure out from environment (default),
 - 1: Use RTE script,
 - 2: Don't use RTE script;

and have this documented in the help section of the `jsonnet`?
It would be something similar to https://github.com/DUNE-DAQ/daqconf/blob/develop/schema/daqconf/confgen.jsonnet#L38 (enum) or https://github.com/DUNE-DAQ/daqconf/blob/develop/schema/daqconf/confgen.jsonnet#L74 (int).

JonathanHancock0:
I've made a new commit with the three options

plasorak (approved):
LGTM

PR #237, opened 2022-10-20
Jhancock/234 configurable rte script usage

plasorak:
Could also go in the patch release


wesketchum (approved):
Looks ok to me, and is working in the integration area (without K8s)

PR #241, opened 2022-11-04
allow env var and tilde in the thread pinning script path

plasorak:


wesketchum (approved):
Looks good, and works at EHN1.

PR #248, opened 2022-11-14
Allow setting the Latency Buffer NUMA conf on a per-RU basis. Allow for

eflumerf:
overriding the FELIX card ID in the FelixCardReader configuration.


eflumerf:
Just noticed that I left some debug printouts in there...

wesketchum (approved):
Confirmed that configs are produced as necessary and can run at EHN1

PR #252, opened 2022-11-16
Fix DQM configuration for multiple APAs

jmcarcell:
The nested loop was wrong and the source ids for the first APA were not being passed while for the second APA they were being duplicated


wesketchum (approved):
Looks good. more importantly, works!

PR #257, opened 2022-11-16
patch to generate appropriate firmware tp configurations

ShyamB97:
- numa aware buffers for tp links
 - updated device names for the dtpcontrollers


wesketchum (approved):
Looks ok to me. Works at EHN1.

PR #259, opened 2022-11-16
Additional trigger algorithm parameters for configuring cold box runs

CharlieBatchelor:
All that has been done here is to include additional parameters to the schema for DUNEDAQ configurations. The parameters in question configure the tried and tested `HorizontalMuon` algorithm to trigger in three different ways.

1. ADC Threshold - When the ADC integral of the current window is over a configurable `adc_threshold`
2. Multiplicity Threshold - When the number of unique channels with hits within the current window exceeds the `n_channels_threshold`
3. Adjacency Threshold - When the hits within a current window form a track of at least `adjacency_threshold` in length (in collection wires!)

There are also flags telling us which of these three methods to use for triggering. Any combination of them can be used simultaneously. The default is to use adjacency triggering.


CharlieBatchelor:
Closing for new #262

Issue #256, opened 2022-11-16
Introduce DAPHNE Streaming mode


ikatza:
The UID for queues: `pds_list_...` 

should be: `pds_stream_`

PR #261, opened 2022-11-17
Change the FRONTEND_TYPE of PDS from pds_list to pds_stream

ikatza:
Fixes issue #256


wesketchum (approved):
Looks good, working in EHN1.

PR #269, opened 2023-01-16
Initial commit

JonathanHancock0:
Adds an option to the conf ("use_data_network" false by default) saying whether to use the control_to_data_network function to find hostnames. Tested that nanorc runs correctly with the option set to true or false, and that config generation is aborted when both use_data_network and use_k8s are true.


plasorak (approved):
LGTM, thanks.

PR #270, opened 2023-01-16
TPZipper cardinality setting bugfix

ArturSztuc:
The TPZipper cardinality was being set to the number of hardware links rather than the number of the TPSet queues during the config generation. This bugfix fixes that.

More info in  [DUNE-DAQ/trigger issue 167](https://github.com/DUNE-DAQ/trigger/issues/167).

### DIFF
--- a/python/daqconf/apps/trigger_gen.py
+++ b/python/daqconf/apps/trigger_gen.py
@@ -217,7 +217,7 @@ def get_trigger_app(CLOCK_SPEED_HZ: int = 50_000_000,
                 # tazipper.max_latency_ms, everything should be fine.
                 modules += [DAQModule(name = f'zip_{region_id}',
                                       plugin = 'TPZipper',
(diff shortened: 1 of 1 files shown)


ArturSztuc:
The bugfix was tested by running
`daqconf_multiru_gen -c ./np04_daq_fwtp_noise.json --hardware-map-file ./hw_maps/np04_hw_map_APA2.txt arturs_test`  on np04 and checking the TPZipper cardinality in the generated config files. 
The results was 2, as expected for that FW TPG run.
Running this on 10 links but with software TPG still results in cardinality of 10, as required.

CharlieBatchelor (approved):
Artur and I went through these bug fix tests together by generating a FW TPG np04 daq configuration which specifically has _x_ lines in the hardware map file, and _y_ queues generated by the `readout` application, where _x_ is not necessarily equal to _y_. In fact, the hardware map in this case had 10 lines, the readout app generated 2 queues, and the expected behaviour for a fix was observed.

PR #274, opened 2023-01-19
Sbhuller/tpset sourceid

ShyamB97:
Adds missing fixes to the tpset source id which were made after patch-v3.2.x was merged into develop.


python/daqconf/apps/readout_gen.py
```diff
@@ -239,7 +239,9 @@ def get_readout_app(DRO_CONFIG=None,
                                           channel_map_name = TPG_CHANNEL_MAP,
                                           emulator_mode = EMULATOR_MODE,
                                           error_counter_threshold=100,
-                                          error_reset_freq=10000
+                                          error_reset_freq=10000,
+                                          tpset_topic="TPSets",
```
eflumerf: Topics are no longer set, is this a holdover?

ShyamB97: ah yes I think so, removed

eflumerf (approved):
LGTM

PR #275, opened 2023-01-19
Fix cardinality configuration changes

CharlieBatchelor:
When we initially fixed the `cardinality` configuration, we forgot to take into account the tests for multiple readout applications, which divides the total number of input TP queues presented to the various `TPZippers`. This amendment accounts for that, and passes the `3ru_3df` integration tests provided in `dfmodules`.


ArturSztuc (approved):
Might be worth noting that we assume len(TA_SOURCE_IDS) is always the same as the number of readout apps.

This was tested at Fermilab, cern's lxplus and np04 resources.

PR #276, opened 2023-01-23
Add boot.connectivity_service_interval configuration parameter to

eflumerf:
control how often connections are re-published to ConnetivityService.

Follows https://github.com/DUNE-DAQ/iomanager/pull/35


ron003 (approved):
tested at np04

PR #279, opened 2023-01-26
Adding in incompleteness tolerance parameters for the trigger zipper.

CharlieBatchelor:
Partnered with [179](https://github.com/DUNE-DAQ/trigger/pull/179) of trigger


wesketchum (approved):
tested and working in ehn1

PR #283, opened 2023-02-02
gcrone/start connectivity service

gcrone:


plasorak:
Just a question, there's no way this connectivity service would look like a "normal app" in the configuration, that gets fed blank json?

gcrone:
The connectivity server won't send the completed messages at the end of a command so sending it a blank json would get an ack but then nanorc would timeout waiting for the response.

plasorak:
gotcha, thanks

plasorak (approved):
Same comment as https://github.com/DUNE-DAQ/nanorc/pull/175

PR #285, opened 2023-02-02
Adding readout map to MLT. It specifies readout pre/post times (ticks…

MRiganSUSX:
…) for individual TC types.


MRiganSUSX:
This is now superseded by PR #305

PR #292, opened 2023-02-22
Allow overridding the data files (frames.bin) to use based on detector

eflumerf:
ID, using the asset manager to resolve all files.


plasorak (approved):
This is pretty cool as it allows us to run in a mixed detector emulator mode.

PR #298, opened 2023-03-03
Removed the TriggerRecordBuilder reply_connection_name configuration parameter

bieryAtFnal:
…from all of the relevant scripts.  Now that we are storing this connection name at INIT time in th TRB code.

This is part of the easy fixes that we are doing as part of the ConnSvc consolidation.

This PR should be tested and review in conjunction with one in the dfmodules repo.


eflumerf (approved):
LGTM integration tests appear to be OK

PR #300, opened 2023-03-06
remove connections names from conf params and add the corresponding g…

glehmannmiotto:
…raph endpoints.
Conf parameters containing connection names have been removed and the endpoint are added with a logical/global name to the modules graph.
This PR goes together with the corresponding one on trigger.


bieryAtFnal (approved):
LGTM

PR #302, opened 2023-03-08
Removed the ReadoutModel conf-time timesync_connection_name parameter

bieryAtFnal:
… because A) it can be determined at INIT time, and B) because the value that it was being set to was wrong.

The changes in this repo were simply to remove the config param from the confgen scripts.

I noticed these problems when working on improving the handling of connection information as part of the ConnectivityService consolidation.

This PR is tied to a corresponding PR in the readoutlibs repo.


eflumerf (approved):
Builds, integration tests behave normally

PR #304, opened 2023-03-16
Added config param to the Readout confgen to enable the use of "now" as the playback data start time

bieryAtFnal:
That is, this new parameter allows the use of the current wallclock time as the first data timestamp when replaying data from files.

This PR should be tested and reviewed with the companion one in the `readoutlibs` package.  ([number 99](https://github.com/DUNE-DAQ/readoutlibs/pull/99))


glehmannmiotto (approved):
LGTM

PR #311, opened 2023-03-28
update default for clock_speed_hz

strilov:


bieryAtFnal:
This may break existing integtests and instructions that make use of the existing 50 MHz default value.  Does this change need to be made now?  If so, what is the reason?  Thanks

strilov:
Hi Kurt. The idea was that since most (possibly now, all, since the DAPHNE firmware has been integrated with the new timing protocol) systems make use of the new timing protocol (where the clock is always 62.5 MHz), it'd make sense that the default clock_speed_hz values are now 62.5 MHz. This is the first step towards removing that parameter as we drop support for the old timing protocol and old detector RO electronics (e.g. protowib) in the future.

bieryAtFnal:
I have merged the changes in this PR to the ones listed in PR #331 (many were identical), so I'm going to close this PR.

PR #318, opened 2023-03-31
Resolve localhost to hostname

plasorak:


glehmannmiotto (approved):
Tested in:
- completely localhost mode
- partially declared hosts
- with and without connectivity service
- with and without a declared host for the connectivity service
All works as expected. I would even say that with this modification we could set "use_connectivity_service" and "start_connectivity_service" to true by default.

PR #322, opened 2023-04-05
add 10000 to the connectivity service port

plasorak:
15000 is a better port for the connectivity service, hopefully.


bieryAtFnal (approved):
LGTM, though it may be useful to have Ron also review this PR and the associated one in nanorc.

ron003 (approved):
runs fine - conn serv is started on 15000 or port specified in config

PR #325, opened 2023-04-13
fix bug where input config is overwritten if you give it full path

wesketchum:


bieryAtFnal (approved):
I've confirmed that the change in this PR avoids the original daqconf.json file getting over-written when the full path is given to the daqconf_multiru_gen command.

bieryAtFnal:
This fix looks good to me, so I'm going to go ahead and merge it, in the interest of finishing things up.  It would be great if @MRiganSUSX could still test it, as confirmation. Thanks

MRiganSUSX:
I have tested this and can confirm that it fixed the issue.

Tested providing the config to daqconf_multiru_gen as:
- full path
- dynamic path
- copied config to the 'run' directory
These all created configurations inside the provided config dir name, and did not modify the configuration file that was passed.

PR #326, opened 2023-04-13
Updated the ru_name that is passed to get_dqm_app in daqconf_multiru_…

bieryAtFnal:
…gen to have underscores in the hostname.

I'm hopeful that this change will fix the problems that Giovanna and Wes have seen with the readout DQM applications at np04 complaining about invalid timestamp.s

### DIFF
--- a/scripts/daqconf_multiru_gen
+++ b/scripts/daqconf_multiru_gen
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+!/usr/bin/env python3
 import click
 import math
 from rich.console import Console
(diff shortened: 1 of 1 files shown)


scripts/daqconf_multiru_gen
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+!/usr/bin/env python3
```
wesketchum: is this right?

wesketchum (changes requested):
other than first line needing the `#` back, looks good / works.

jcfreeman2 (approved):
LGTM

PR #329, opened 2023-04-20
daqconf_multi_ru `file-label` option to set the file prefix

### LINKED ISSUE #328: Decoupling raw data file name prefix from operational environment.
The operational environment attribute is currently used to set the hdf5 raw data file prefix.
The operational environment is supposed to be used to identify the apparatus used to take data. 
On several occasions, during specific tests, the operational environment was used to set the raw data file prefix instead (e.g. `longreadout`, `swtpg`, `fwtpg`) etc.
When used in this way, the operational environment value hdf5 attributes cannot be used to tell what experiment the data was collected with, which in turn is problematic for data analysis tools that may need to know the experiment id to apply the correct settings (I.e. `justintime` & tpc channel map).
To avoid such mix, it would be better to 
1) have a separate parameter for setting the datawriter file prefix
2) change the `daqconf_multiru` commandline to set the prefix (or label) and not the op_env attribute (which is part of boot.json)
3) default `file_label` to `op_env` if no value is provided

alessandrothea:
Solves issue described in #328.
Closes #328


plasorak:
Shouldn't the file-label go into the dataflow schema?

bieryAtFnal:
@plasorak , my understanding is that the changes in this PR allow us to specify the raw data filename on the daqconf_multiru_gen command line.  There is already a file_prefix configuration parameter in the dataflow schema, and these changes allow us to set that parameter independently from the operational environment (which can still be set from a daqconf.json "boot" param);

@alessandrothea , I noticed a slight omission in the dataflow_gen.py script, which I will push to this branch momentarily.  With that change, my testing looks good.

bieryAtFnal (approved):
LGTM

plasorak:
Good for me too

PR #330, opened 2023-04-20
Use the configuration objects to configure application

plasorak:
The basic idea here is to move application-specific code from `daqconf_multiru_gen` into their own generators in `python/apps`. Still a draft.


alessandrothea:
I think that's the (ultimate) goal. For the first iteration, a good (and achievable) goal would be to 
1) review and refactor the parameters in the schema to clarify the context (app specific or global)
2) make use of the structures in the generator functions

PR #333, opened 2023-04-26
Resolve IP only when not running with K8s

plasorak:
Related PR: https://github.com/DUNE-DAQ/iomanager/pull/55
Issue: https://github.com/DUNE-DAQ/iomanager/issues/50

Sets an env variable to 0 when we don't want to have hostname->IP in iomanager/ipm


eflumerf:
Other changes in ipm/iomanager (See https://github.com/DUNE-DAQ/ipm/pull/86) make this change unneeded.

PR #335, opened 2023-05-18
removing different streambuffer size for pacman

jtenavidal:
In the current readout configuration, the stream buffer size is set to `stream_buffer_size = 8388608`, with the exception of pacman. For pacman it was set to a 100. 

After discussions with @willpvazquez we concluded that there was no specific reason to have a different buffer size for pacman. 

In this pull request, I use the same stream buffer size for pacman, and the integration tests are successful. I configured my environment with N23-05-15.

@alessandrothea can introduce the same changes to his refactorization of readout modules.


jtenavidal:
Integration test output: 
`..Sanity-check passed
Event count is within a tolerance of 10 from an expected value of 60
PACMAN fragment count of 1 confirmed in all 60 events
.

=================================================================================== 3 passed in 132.85s (0:02:12) ==========`

willpvazquez (approved):

plasorak:
Eugh, I have not clue what this is supposed to be, I'm happy for you to merge

jtenavidal:
> Eugh, I have not clue what this is supposed to be, I'm happy for you to merge

HI Pierre, 
Alessandro and I were discussing if we needed a specific setup for the streambuffer size for pacman. We concluded there's no real reason to keep it that way.  I tagged you as well, so you are aware of the change.

bieryAtFnal (approved):
LGTM, test_pacman-raw.py and other integration tests continue to run successfully.

PR #337, opened 2023-05-24
Introducing the Detector Readout map 

### LINKED ISSUE #327: Readout application generator refactoring
The readout application generator function has grown organically over time to the point where making (any) modifications has become very error prone. 
The code needs to be reviewed and refactored. The refactored code should clearly reflect the structure of the readout application and highlight its key components: card readers, data link handlers, data processing and utilities.

### LINKED ISSUE #336: Introducing the Detector-Readout Stream map
The transition to Ethernet readout introduces a different kind of connection between detector electronics and readout.
The `HardwareMap` definition is FELIX centric, describing the readout end in terms of FELIX link hardware coordinates (host, card id, SLR, link). Ethernet connection description requires a different set of parameters that cannot be fit in the existing one.
While the option of adding new columns to the existing hardware map would work, it would result in a quite cumbersome map file to edit. To circumvent these limitations, the `DetectorReadoutMap` is introduced.

The `DetectorReadoutMap` is similar to the `HardwareMap` in mapping link (or stream) ids to geo ids. It differs from `HardwareMap` in how it handles the connection hardware parameters, which are grouped in a separate record and can change from stream to stream based on the stream kind. Also, the map file format is json.

Example of a Detector-Readout Stream map entry:

```json
    {
        "src_id": 100,
        "geo_id": {
            "det_id": 2,
            "crate_id": 1,
            "slot_id": 0,
            "stream_id": 0
        },
        "tech": "flx",
        "config": {
            "protocol": "half",
            "mode": "var_rate",
            "host": "np04-srv-030",
            "card": 0,
            "slr": 0,
            "link": 0
        }
    },
```

alessandrothea:
Covers #336 and #327


eflumerf (approved):
LGTM, builds, unit tests pass, basic confgen appears to work.

PR #338, opened 2023-05-24
Thea/dro map fix

eflumerf:


bieryAtFnal (approved):
These changes helped me generate a configuration that worked.

PR #342, opened 2023-05-31
check value of tp_mode argument in generate_trigger_source_ids

bieryAtFnal:
so that we don't try to generate TP sources when TPs are not enabled....


glehmannmiotto (approved):
LGTM

PR #343, opened 2023-06-01
Directory mounting frenzy

plasorak:


plasorak:
If we are not mounting the install directory, we don't need to have an entrypoint.

plasorak:
Closing this as not the correct way to do things.

PR #344, opened 2023-06-01
Restored the ability to set the emulator mode config param for RawDat…

bieryAtFnal:
…aProcessorConf when TPs are not enabled

I'm hopeful that this fixes the immediate problem of the readout.emulator_mode flag not getting set correctly, but I'm concerned that the newest readout_gen.py misses other config params when configurating RawDataProcessors (when TPG is not enabled).

### DIFF
--- a/python/daqconf/apps/readout_gen.py
+++ b/python/daqconf/apps/readout_gen.py
@@ -467,7 +467,8 @@ def create_det_dhl(
         DATA_REQUEST_TIMEOUT: int,
         FRAGMENT_SEND_TIMEOUT: int,
         RAW_RECORDING_ENABLED: bool,
-        RU_DESCRIPTOR # ReadoutUnitDescriptor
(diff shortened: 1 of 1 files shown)


glehmannmiotto (approved):
This will fix the emulation flag, indeed, but I agree with @bieryAtFnal that the idea of generating a "default" raw processor conf and overriding its settings only then TPs are enabled does not seem right. It could make sense if there was a different organisation of the schema, grouping only TPG parameters together, but the raw processor is not only hit finding. It is also doing pre-checks before inserting data into the latency buffer (and doing some of the emulation tasks).

bieryAtFnal:
I'm especially concerned about whether the clock frequency combined with the slowdown factor needs to be added back..   We'll see...

I'm going to go ahead and merge this change, but we should be alert to additional changes that will be needed in this section of code.  (and if the author of this section of readout_gen.py has a chance to look at the design with Giovanna's comments in mind, that would be great).

PR #350, opened 2023-06-14
Updated the tpg_algorithm default in confgen.jsonnet to be SimpleThre…

bieryAtFnal:
…shold since the existing default of SWTPG is no longer available


roland-sipos (approved):
LGTM

PR #351, opened 2023-06-16
Eflumerf/expand vars for data files

eflumerf:
Expand environment variables in the data file specifications (e.g. "default_data_file"), so that `"default_data_file": "$PWD/frames.bin"` works as expected.


plasorak (approved):
LGTM

PR #352, opened 2023-06-17
Conf for ERSStreamer

mroda88:


mroda88:
Already tested and working, but first we need to setup all the rest of the infrastructure

PR #355, opened 2023-06-26
Jhancock/daqconf viewer

JonathanHancock0:
By running the "daqconf_viewer" command, the user can open a graphical view of all the configurations in the database. They can also look at a "diff" of two configs, which has similar formatting to the one used on github.

### DIFF
--- a/scripts/daqconf_viewer
+++ b/scripts/daqconf_viewer
@@ -6,19 +6,23 @@ import json
 import sys
 
 from rich.text import Text
+from rich.markdown import Markdown
 from difflib import context_diff, ndiff, unified_diff
 
 from textual import log, events
 from textual.app import App, ComposeResult
 from textual.containers import Horizontal, Content, Container, Vertical
 from textual.widget import Widget
-from textual.widgets import Button, Header, Footer, Static, Input, Label, ListView, ListItem, Tree
+from textual.widgets import Button, Header, Footer, Static, Label, ListView, ListItem, Tree
 from textual.reactive import reactive, Reactive
-from textual.message import Message, MessageTarget
 from textual.screen import Screen
 
 auth = ("fooUsr", "barPass")
 oldconf = None
+
+class TitleBox(Static):
+    def __init__(self, title, **kwargs):
+        super().__init__(Markdown(f'# {title}'))
 
 class LabelItem(ListItem):
     def __init__(self, label: str) -> None:
@@ -39,6 +43,7 @@ class Configs(Static):
         self.set_interval(0.1, self.update_configs)
 
     def compose(self) -> ComposeResult:
+       yield TitleBox('Configuration List')
        yield ListView(LabelItem("asfas"))
 
     async def update_configs(self) -> None:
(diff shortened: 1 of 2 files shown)


scripts/daqconf_viewer
```diff
@@ -0,0 +1,261 @@
...
+    def compose(self) -> ComposeResult:
+        yield Static(id='diffbox')
+
+    async def get_json(self, conf, ver) -> None:
+        self.confname = conf
+        self.version = ver
+        if self.confname != None and self.version != None:
+            async with httpx.AsyncClient() as client:
+                payload = {'name': self.confname, 'version': self.version}
+                r = await client.get(f'{self.hostname}/retrieveVersion', auth=auth, params=payload, timeout=60)
```
plasorak: I probably would make this timeout 2seconds or so, and then make
```
r = {"error": f"couldn't retrieve the configuration at {self.hostname}/retrieveVersion (payload: {payload}"}
```
and wrap in in a try-except

scripts/daqconf_viewer
```diff
@@ -0,0 +1,261 @@
...
+    def compose(self) -> ComposeResult:
+       yield Tree("", id='conftree')
+
+    async def get_json(self, conf, ver) -> None:
+        self.confname = conf
+        self.version = ver
+        if self.confname != None and self.version != None:
+            async with httpx.AsyncClient() as client:
+                payload = {'name': self.confname, 'version': self.version}
+                r = await client.get(f'{self.hostname}/retrieveVersion', auth=auth, params=payload, timeout=60)
```
plasorak: same comment as below (quicker timeout and try/except)

scripts/daqconf_viewer
```diff
@@ -0,0 +1,261 @@
...
+
+    def on_mount(self) -> None:
+        self.set_interval(0.1, self.update_configs)
+
+    def compose(self) -> ComposeResult:
+       yield ListView(LabelItem("asfas"))
+
+    async def update_configs(self) -> None:
+        async with httpx.AsyncClient() as client:
+            r = await client.get(f'{self.hostname}/listConfigs', auth=auth, timeout=60)
```
plasorak: and here too

scripts/daqconf_viewer
```diff
@@ -0,0 +1,261 @@
...
+
+    def action_make_diff(self):
+        '''Saves the current config to a global variable, then pushes the diff screen.'''
+        dis = self.query_one(Display)
+        if dis.confdata != None:
+           global oldconf
+           oldconf = dis.confdata
+           self.push_screen('diff')
+
+if __name__ == "__main__":
```
plasorak: Could you import click and make a `--help` function here? We also may want to use a different hostname (the one in line 238) for the mongo db, it would be nice to be able to specify it in the CLI.

scripts/daqconf_viewer
```diff
@@ -0,0 +1,261 @@
...
+        self.set_interval(0.1, self.update_versions)
+
+    def new_conf(self, conf) -> None:
+        self.current_conf = conf
+
+    async def update_versions(self) -> None:
+        if self.current_conf:
+            async with httpx.AsyncClient() as client:
+                payload = {'name': self.current_conf}
+                r = await client.get(f'{self.hostname}/listVersions', auth=auth, params=payload, timeout=60)
```
plasorak: same comment for the try cathc

plasorak (commented):
It would be nice to have title at the top of the boxes (available configurations, version and configuration data), but otherwise modulo minor comments below, this is looking very good and works well.

scripts/daqconf_viewer
JonathanHancock0: Since this is a different widget, I did this by having the except block call a function that finds the display and puts the appropriate text there.

scripts/daqconf_viewer
JonathanHancock0: Same here.

PR #356, opened 2023-06-27
MLT: Coincidence trigger (trigger deliverable for v4.1.0)

MRiganSUSX:
This is the implementation of coincidence trigger as discussed in
[Trigger: issue 220](https://github.com/DUNE-DAQ/trigger/issues/220)


python/daqconf/apps/trigger_gen.py
```diff
@@ -70,6 +72,23 @@ def get_buffer_conf(source_id, data_request_timeout):
...
+        tmp_bits = []
+        for bit_name in bitword:
+            bit_value = trgbs.string_to_fragment_type_value(bit_name)
+            if bit_value == -1:
+                raise RuntimeError(f'One or more of provided MLT trigger bitwords is incorrect! Please recheck the names...')
+            else:
+                tmp_bits.append(bit_value)
+        final_bit_flags.append(tmp_bits)
+ 
+    return final_bit_flags
```
ArturSztuc, line 91: I was confused by the variable naming for a moment.
Maybe it would be better to name them something like
`bitwords` -> `algo_combinations`
`bitword` -> `algo_names`
`bit_name` -> `algo_name`

(or something better than that...) since these are not bitwords yet.

config/daqconf_full_config.json
```diff
@@ -235,6 +235,8 @@
                 "time_before": 1000
             }
         },
+        "mlt_trigger_bitwords": [],
+        "mlt_use_bitwords": false,
```
ArturSztuc, line 239: I'm wondering if it would be possible to make it so the `mlt_use_bitwords` is implied to be false anyway if the `mlt_trigger_bitwords` is empty, and true if not empty, so we wouldn't need it altogether (so it would be implied inside of the `get_trigger_app function`).
At the same time maybe it's better to have an explicit flag anyway? Meh.

ArturSztuc (approved):
Looks good! I made a couple comments, but not significant -- so please ignore them until after the PR. I don't see anything that would cause any problems here.

config/daqconf_full_config.json
MRiganSUSX: As discussed, keeping the flat gives more flexibility, such as testing with the flag on but no bitwords - ie no trigger decisions. Keeping as is for now.

python/daqconf/apps/trigger_gen.py
MRiganSUSX: hmm, the names could probably be improved, although I find your suggested naming more misleading - as we are not necessarily dealing with algos. Will skip for now and we can revisit later.

PR #366, opened 2023-07-06
Put back missing comparison for det_str in readout_gen compute_data_types

bieryAtFnal:
This bug seems to have crept in as part of the DAPHNE data type changes yesterday.

I noticed it when I ran the daq-systemtest/integtest/readout_type_scan, and the TDE part of that test failed because there were no TDE_AMC fragments found.


eflumerf (approved):
LGTM

PR #367, opened 2023-07-07
Updated version to v6.0.1

bieryAtFnal:


eflumerf (approved):
Version appears to be updated.

PR #373, opened 2023-07-13
Get rid of the entrypoint

plasorak:
For K8s, since we are now mounting DBT_INSTALL_DIR, we can now directly use `daq_app_rte.sh` rather than having an entry point.


plasorak:
Works with: https://github.com/DUNE-DAQ/nanorc/pull/208

dingp (approved):
LGTM

plasorak:
This PR generates configurations which always have the RTE script (which it figures out from the local environment), and therefore gets rid of the RTE flag in the boot configuration. Also, it removes the entry point which will now be wrong since we are all running off of CVMFS.

PR #376, opened 2023-07-26
Jhancock/add run registry

JonathanHancock0:
Added a new screen to the app that acts as an interface for the run-registry microservice. The user can look at the metadata for any run, and browse the configs in a similar format to the other screens.


plasorak (commented):
Looks good, and works. Is it possible to diff configurations from the file system on in the run registry?

PR #378, opened 2023-08-07
Remove ADC generation from FrameProcessor

adam-abed-abud:
Removed ADC generation from FrameProcessors (pre-processing task) and added instead the ADC pattern generator in the SourceEmulatorModel.

A new DAQConf entry has been added for enabling the ADC pattern generation.

This fixes issues https://github.com/DUNE-DAQ/fdreadoutlibs/issues/110 and https://github.com/DUNE-DAQ/fdreadoutlibs/issues/124


bieryAtFnal (approved):
LGTM

PR #379, opened 2023-08-24
K8S readout consolidation

alessandrothea:
Moving container attributes for readout from nanorc to daconf.
Also, removes obsolete ND code


plasorak (approved):
LGTM, I'm curious to see what the capabilities do.

plasorak (approved):
LGTM, allows us to use any host when use_fake_card is true.

PR #380, opened 2023-09-05
Modified the TP_rate_per_ch config param to support fractional values

bieryAtFnal:
I found this useful during testing.

It is correlated with a companion PR in the `readoutlibs` repo, [number 137](https://github.com/DUNE-DAQ/readoutlibs/pull/137).


adam-abed-abud (approved):
LGTM.

PR #385, opened 2023-09-20
Add script to generate OKS Session from boot.json

gcrone:


plasorak (approved):
This has been confirmed to work.

PR #387, opened 2023-09-25
Removed some stale information that contained references to the now-o…

bieryAtFnal:
…bsolete daqconf_multiru_gen script.


bieryAtFnal:
@jcfreeman2  , @plasorak , I'd appreciate your input on whether you think that it is OK to drop this stale information.

plasorak:
Yep looks good.

PR #395, opened 2023-10-10
Gcrone/oks drunc controller conf

gcrone:
This updates the OKS generation script to generate an OKS session compatible with the latest coredal schema.

### DIFF
--- a/scripts/generate_oks_session
+++ b/scripts/generate_oks_session
@@ -87,12 +87,45 @@ def generate(bootfile, datafile, session_name):
   # Create a run controller for the segment. We say it runs on the last
   # specified virtual host for now since this is not defined in the json
   # ###########################################################################
+  ifname = 'user-providedirun-number'
+  iface = dal.FSMinterface('dummy-if', name=ifname, commands=[])
+  db.update_dal(iface)
+  pre_trans = dal.FSMxTransition("start", order=[ifname], mandatory=[ifname])
+  db.update_dal(pre_trans)
+  transitions = []
+  trans_conf = [
+    ("conf", "initial", "configured"),
+    ("start", "configured", "ready"),
+    ("enable_triggers", "ready", "running"),
+    ("disable_triggers", "running", "ready"),
+    ("drain_dataflow", "ready", "dataflow_drained"),
+    ("stop_trigger_sources", "dataflow_drained", "trigger_sources_stopped"),
+    ("stop", "trigger_sources_stopped", "configured"),
+    ("scrap", "configured", "initial")
+  ]
+  for trig,src,dest in trans_conf:
+    trans = dal.FSMtransition(trig, source=src, dest=dest)
+    db.update_dal(trans)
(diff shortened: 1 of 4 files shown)


scripts/generate_oks_session
```diff
@@ -84,15 +89,70 @@ def generate(bootfile, datafile, session_name):
     vhostdals[h] = vhdal
 
   # ###########################################################################
-  # Create a run controller for the segment. We say it runs on the last
-  # specified virtual host for now since this is not defined in the json
+  # Create a run controller for the segment. 
   # ###########################################################################
+  fsm = controller['statefulnode']['fsm']
+  for iface in fsm['interfaces']:
+    ifname = 'user-providedirun-number'
```
plasorak: user-providedirun-number -> user-provided-run-number

scripts/generate_oks_session
```diff
@@ -84,15 +89,70 @@ def generate(bootfile, datafile, session_name):
...
+                             post_transitions = post_trans,
+                             command_sequences=cmd_sequences)
+  db.update_dal(fsm)
+  bcaster = dal.RCBroadcatser('bcaster',
+                              type=controller['broadcaster']['type'],
+                              address=controller['broadcaster']['kafka_address'],
+                              publish_timeout=controller['broadcaster']['publish_timeout'])
+  db.update_dal(bcaster)
   name = "my-controller"
   args = f"--conf-type OKS {datafile} ${{PORT}} {name}"
```
plasorak: Not strictly related to this PR, but this is not how it is done in the [drunc controller code](https://github.com/DUNE-DAQ/drunc/blob/develop/src/drunc/controller/configuration.py#L10C14-L10C22). Right now, we use URIs. So that would look like: `oks:///path/to/oks/file.xml`.

plasorak (commented):
2minor comments.

scripts/generate_oks_session
plasorak: Actually, you don't need this one right? its filled automatically

gcrone: Yeah, I think that is a mistake leftover from when I was just inventing a configuration on the fly rather than converting from your controller.json. I will fix this

scripts/generate_oks_session
gcrone: Yes, I changed that this morning, haven't pushed my update yet but I changed it to `oksconfig:{cfg_path}` (cfg_path only starts with single '/') but only if you give an extra option `--oks_controller`. It now defaults to `file://{ctrlr_path}` (ctrlr_path is the abolute path to controller.jso). This allows drun to start a controller configured by JSON rather than OKS while we haven't written the OKS parsing for the controller.

plasorak: Very good.

gcrone:
I think the changes I just pushed should deal with your comments so far.

plasorak (approved):
LGTM

PR #396, opened 2023-10-11
Mrigan/hsi config fix

MRiganSUSX:
This is a PR for [trigger issue 239](https://github.com/DUNE-DAQ/trigger/issues/239). More details there.

### DIFF
--- a/python/daqconf/apps/readout_gen.py
+++ b/python/daqconf/apps/readout_gen.py
@@ -284,6 +284,7 @@ class ReadoutAppGenerator:
             mgraph: ModuleGraph,
             dlh_list: list,
             RUIDX: str,
+            GEN_TIMESYNC: bool,
         ) -> None: 
         """Adds detector readout endpoints and fragment producers"""
         for dlh in dlh_list:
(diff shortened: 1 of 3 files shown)


bieryAtFnal (approved):
The changes look fine to me, and I see no problems when running integtests.

It took me a few minutes to understand some of the changes, though.

- I'm guessing that the addition of the test on the presence of one or more TC sources is just a nice sanity check now that there isn't *always* an HSI input configured.  And, it doesn't have anything to do with ensuring that there are receivers for TimeSync messages.
- And, I'm guessing that the addition of the Tees on the outputs of the RTCM and CTCM also doesn't have anything to do with the HSI input, but rather was something that should have been included earlier and is being added now for completeness.

If I've guessed wrong on either of those, it might be nice to chat briefly about them, but I don't feel that this should hold up the PR.

MRiganSUSX:
Sorry for late answer. You are correct on both points. Perhaps the second one is a little more than just completeness - these are actually required for trigger to function properly in cases when there is no timing trigger candidate maker (there are some daqconf checks on creation of queues, and this makes them pass)

ArturSztuc (commented):
Looks good, doesn't look like there are any changes that would break backward compatibility.

PR #399, opened 2023-10-12
Mroda/kluster

mroda88:


mroda88:
This is the change necessary to use the new daq-kube cluster at np04.

mroda88:
Not everything is working on the cluster side, so it needs more work and investigation. 

Also, so far I tested the monitoring part of the cluster, experts should check other repos like run-number, and connectivity service

mroda88:
Since the plan is to switch the DNS record, no changes will be needed on daqconf. I'll downgrade this PR as a draft for testing purposes. When the switch is done, we'll close the PR and delete the branch.

mroda88:
DNS has been changed and so this becomes obsolete. Closing it.

PR #404, opened 2023-10-31
Attempting to clean up the presence, absence, and initialization of the source_id config parameter in various readout modules

bieryAtFnal:
This PR is coupled with a companion one in the `readoutlibs` repo, so they should be tested and merged together.

I'm going to start out with this PR as a draft so that we can discuss it before moving forward with it.

- The source_id parameter is proposed to be dropped from the latencybufferconf, since it is not used in that code.
- The value for the source_id parameter is proposed to be set in the rawdataprocessconf since it *is* being used by the code, but is not currently being specified correctly.

[readoutlibs PR 144](https://github.com/DUNE-DAQ/readoutlibs/pull/144)


adam-abed-abud (approved):
Looks good to me.

PR #407, opened 2023-11-02
readout_gen: method to specify custom mounts and resources

alessandrothea:
Required for k8s configurations


bieryAtFnal (approved):
looks reasonable to me

PR #412, opened 2023-11-15
MLT readout map rework

MRiganSUSX:
Fix for custom readout map used in MLT to only apply custom readout windows to explicitly configured TC types. 
Please see details in: [custom_readout_map_fix.pdf](https://github.com/DUNE-DAQ/trigger/files/13377208/custom_readout_map_fix.pdf)
Required: [Trigger PR248](https://github.com/DUNE-DAQ/trigger/pull/248)


ArturSztuc (approved):
LGTM, good reduction!

bieryAtFnal (approved):
LGTM.

I've run tests in which I've separately controlled the readout window for two individual TC types in an emulated-data system in which I have three types of TCs being produced (timing, prescale, random).

As part of this, I made some enhancements to the `HDF5LIBS_TestDumpRecord` test application to print out the request window that was used for each SourceID in a TriggerRecord on disk.  I'll contribute those independently of this PR.

PR #413, opened 2023-11-28
Added a --graceful-timeout=0 option to the connection service boot parameters

bieryAtFnal:
…to get it (and its host gunicorn instance) to shut down in a timely way on certain computers.

I first noticed this issue when running a group of automated integration tests on the daq.fnal.gov computer with the `daqsystemtest_integtest_bundle.sh` script.  A small fraction of the time, the 2nd or 3rd or 4th test would fail to successfully start, and it would complain that a new Connectivity Service instance could not be started because one was already running.  Upon investigation, I found that Connectivity Service instances were always taking 10s of seconds to shut down when each of the individual integtests finished.  However, the observed problem of not being able to start a new instance was only happening when the same ConnSvc port number was randomly assigned to two subsequent tests.

It's straightforward to reproduce the long-shutdown behavior on daq.fnal.gov with a simple set of instructions, and I'll post those instructions in a comment on this PR.  However, there are several oddities the I observed while trying to debug this.

One oddity is that other computers don't have the same long delay.  I'll post test results in a separate comment on this PR.  I wonder if there are different OS/system settings on different computers, and that is affecting this behavior.

Another oddity is that the `ConnectivityService/gunicorn` instance responds differently to a 'kill' signal depending on whether it is started by `nanorc` or it is started by hand.

A third oddity is that the Python tools that `nanorc` uses to start/stop processes seem to indicate that the ConnSvc process has shut down when it truthfully hasn't.

Independent of all of those unexpected behaviors, when I was doing some reading about `gunicorn` shutdown behavior, I found posts about including `exit(0)` statements at the end of certain code blocks (e.g. [here](https://stackoverflow.com/questions/75664453/how-to-handle-correctly-sigterm-in-a-gunicorn-app)), and posts about the use of the `gunicorn` --graceful-timeout configuration parameter.

I experimentally determined that adding this parameter to our boot.json files in the Connectivity Service startup block with a value of zero helps get the ConnSvc to exit in a timely way on the daq.fnal.gov computer.  This PR includes that change, in case we decide that this is the best way forward, at the current time.

### DIFF
--- a/python/daqconf/core/conf_utils.py
+++ b/python/daqconf/core/conf_utils.py
@@ -744,6 +744,7 @@ def generate_boot(
                     "--workers=1",
                     "--worker-class=gthread",
                     f"--threads={boot_conf.connectivity_service_threads}",
+                    "--graceful-timeout=0",
                     "--timeout=0",
                     "--pid={APP_NAME}_{APP_PORT}.pid",
                     "connection-service.connection-flask:app"


bieryAtFnal:
Here are steps that can be used to investigate how long it takes for the `ConnSvc/gunicorn` instance to shut down at the end of the `nanorc` session:

```
source /cvmfs/dunedaq.opensciencegrid.org/setup_dunedaq.sh
setup_dbt latest
dbt-create -c -n NAFD23-11-28 28NovFDDevConnSvcTest
cd 28NovFDDevConnSvcTest/sourcecode
git clone https://github.com/DUNE-DAQ/daqsystemtest.git -b develop
cd ..

dbt-workarea-env
dbt-build -j 20
dbt-workarea-env

mkdir rundir
cd rundir

cat <<EOF1 > daqconf.json
{
  "boot": {
    "use_connectivity_service": true,
    "start_connectivity_service": true,
    "connectivity_service_host": "localhost",
    "connectivity_service_port": 15432
  }, 
  "daq_common": {
    "data_rate_slowdown_factor": 1
  },
  "detector": {
    "clock_speed_hz": 62500000
  },
  "readout": {
    "use_fake_cards": true,
    "default_data_file": "asset://?label=WIBEth&subsystem=readout"
  },
  "trigger": {
    "trigger_window_before_ticks": 1000,
    "trigger_window_after_ticks": 1000
  },
  "hsi": {
    "random_trigger_rate_hz": 1.0
  }
}
EOF1

cat <<EOF2 > dro_map.json
[
    {
        "src_id": 100,
        "geo_id": {
            "det_id": 3,
            "crate_id": 1,
            "slot_id": 0,
            "stream_id": 0
        },
        "kind": "eth",
        "parameters": {
            "protocol": "udp",
            "mode": "fix_rate",
            "rx_iface": 0,
            "rx_host": "localhost",
            "rx_mac": "00:00:00:00:00:00",
            "rx_ip": "0.0.0.0",
            "tx_host": "localhost",
            "tx_mac": "00:00:00:00:00:00",
            "tx_ip": "0.0.0.0"
        }
    },
    {
        "src_id": 101,
        "geo_id": {
            "det_id": 3,
            "crate_id": 1,
            "slot_id": 0,
            "stream_id": 1
        },
        "kind": "eth",
        "parameters": {
            "protocol": "udp",
            "mode": "fix_rate",
            "rx_iface": 0,
            "rx_host": "localhost",
            "rx_mac": "00:00:00:00:00:00",
            "rx_ip": "0.0.0.0",
            "tx_host": "localhost",
            "tx_mac": "00:00:00:00:00:00",
            "tx_ip": "0.0.0.0"
        }
    }
]
EOF2

fddaqconf_gen -c ./daqconf.json --detector-readout-map-file ./dro_map.json mdapp_wibeth

nanorc --partition-number 5 mdapp_wibeth ${USER}-test boot wait 3 terminate

let loop_count=1
while [[ 1 ]] ; do
  date
  ps -fu ${USER} | grep -i gunicorn
  sleep ${loop_count}
  let loop_count=${loop_count}+1
done
```

bieryAtFnal:
Here are some results from measuring the `ConnSvc/gunicorn` shutdown times before and after the --graceful-timeout parameter was added to the ConnSvc startup...

```
Computer          Shutdown time before      Shutdown time after
----------        ----------------------    --------------------
daq.fnal.gov          ~30 seconds                 0 (zero)
iceberg03             ~30 seconds                 0 (zero)
np04-srv-024           ~1 second                  0 (zero)
np04-srv-004            0 (zero)                  0 (zero)
```

bieryAtFnal:
I described the status of this investigation at the CCM meeting earlier today, and that group was OK with this change.  I also mentioned it to Alessandro, who is also OK with it.

So, in the spirit of improving the current situation without spending more time investigating code that will become obsolete, I'm going to go ahead and merge this change.

PR #414, opened 2023-12-06
Configurable send partial fragments in daq conf

jtenavidal:
This pull requests fixes [this issue](https://github.com/DUNE-DAQ/lbrulibs/issues/64) 
Altogether with

- https://github.com/DUNE-DAQ/daqconf/pull/414
- https://github.com/DUNE-DAQ/nddaqconf/pull/12
- https://github.com/DUNE-DAQ/lbrulibs/pull/65

In daqconf, the only change is to add an additional optional parameter in the generate(..) function, so that send_partial_fragments can be set to true. By default, the argument is set to false. Therefore, it does not affect the behaviour of other systems. This parameter can be configured from the integration tests readout options.

### DIFF
--- a/python/daqconf/apps/readout_gen.py
+++ b/python/daqconf/apps/readout_gen.py
@@ -386,7 +386,6 @@ class ReadoutAppGenerator:
             SOURCEID_BROKER,
             data_file_map,
             data_timeout_requests,
-            send_partial_fragments=False
             ):
         """Generate the readout applicaton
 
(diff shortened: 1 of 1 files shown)


jtenavidal:
Build instructions available https://github.com/DUNE-DAQ/daqconf/wiki/Instructions-for-setting-up-an-ND-development-software-area

bieryAtFnal (approved):
LGTM.
I ran the test_mpd-raw.py integtest with this change and the corresponding ones in the nddaqconf and lbrulibs repos, and that ran fine.

PR #423, opened 2024-02-13
Extended pcie address regex to handle hex values

wesketchum:
Tested at EHN1 and it works. This PR is for the prep-release branch.


bieryAtFnal (approved):
LGTM

PR #425, opened 2024-02-15
Attempted to disable the creation of TP-related modules and connectio…

wesketchum:
…ns for non-WIBEth readout types

Tested at EHN1 and it seems to work. I'd request that it be included in the 4.3.0 release.

However, I'll note, that if I understand the code it's just assuming that only eth readout will be used for TP generation. At some point we may want to have DAPHNE felix-based readout also do TP generation, while still having other readout applications (e.g. CRT) not do so.

Beyond this, I think there's a bigger general issue here. It doesn't make a lot of sense to me that readout applications that are reading out different hardware should by default be configured in the same way. E.g. latency buffer sizes could well be different, TP algorithms could be different, etc. We can manage these via overrides now, but it would be nice if there could be a better way.

### DIFF
--- a/python/daqconf/apps/readout_gen.py
+++ b/python/daqconf/apps/readout_gen.py
@@ -404,7 +404,7 @@ class ReadoutAppGenerator:
         """
         numa_id, latency_numa, latency_preallocate = self.get_numa_cfg(RU_DESCRIPTOR)
         cfg = self.ro_cfg
-        TPG_ENABLED = cfg.enable_tpg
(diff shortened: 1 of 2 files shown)


bieryAtFnal:
Yes, sorry, I should have said that I/we need to find a more robust way of identifying DAPHNE fragments.  In some of our DRO_map generation code, it is a combination of factors, and I'll look into improving these changes for fddaq-v4.4.0.

bieryAtFnal (approved):
LGTM.  regression tests from the daqsystemtest package work with this change.

PR #431, opened 2024-03-14
Added an offline_data_stream config param to the detector daqconf group and the dataflow datastore parameters

bieryAtFnal:
The changes will allow creators of DAQ configurations to specify the value of the field that we pass to Offline in the file-transfer metadata that indicates the type of data in each run.  (This field is named `data_stream` in the file-transfer metadata, and we have chosen to call it `offline_data_stream` in our code and configurations in an attempt to indicate that it is not used in the DAQ for any purpose.)

This PR is part of a Deliverable for fddaq-v4.4.0, DUNE-DAQ/daq-deliverables#126, and it is coupled with corresponding PRs in other repos, such as DUNE-DAQ/dfmodules#334 and DUNE-DAQ/hdf5libs#75.


andrewmogan (approved):
Tested using Kurt's instructions [here](https://github.com/DUNE-DAQ/daq-deliverables/issues/126).

PR #435, opened 2024-03-19
Asztuc/trigger max tot filtering

ArturSztuc:
Added TP filtering based on TP's time-over-threshold to the Trigger `TPChannelFilter` module, as discussed in https://github.com/DUNE-DAQ/ehn1-operations-issues/issues/11. The default value set to 10,000 ticks, i.e. TPs with TOT larger than 10,000 ticks will not be progressed passed `TPChannelFilter` (but will be still stored in the readout buffers, so can be retrieved as fragments). 

Full descriptions in https://github.com/DUNE-DAQ/trigger/pull/284.


aeoranday (approved):
LGTM. Tested and working with the changes in `trigger`.

PR #436, opened 2024-03-22
New ChannelDistance Algorithm

aeoranday:
With the new ChannelDistance trigger algorithm comes new configurable parameters. This is just to keep up to date with those changes.

This is a result from https://github.com/DUNE-DAQ/triggeralgs/pull/60.


schema/daqconf/triggergen.jsonnet
```diff
@@ -37,6 +37,10 @@ local cs = {
     s.field("adc_threshold", types.count, default=10000),
     s.field("n_channels_threshold", types.count, default=8),
     s.field("print_tp_info", types.flag, default=false),
+    s.field("min_tps", types.count, default=20),
+    s.field("max_time_delta", types.count, default=8000),
```
ArturSztuc: It's very similar to the idea to `window_length`... maybe just use that instead?

ArturSztuc (changes requested):
Need to add these extra lines to the full daq configuration example, here: https://github.com/DUNE-DAQ/daqconf/blob/production/v4/config/daqconf_full_config.json

Otherwise LGTM

schema/daqconf/triggergen.jsonnet
aeoranday: I agree that both are very similar, but I found it confusing because `window_length` implies the window will have a specific length while `max_time_delta` specifies an upper bound to that length.

ArturSztuc: The "window having specific length" is not true for most algorithms, you can set the length to e.g. 10,000, but can still get effectively smaller windows if you reached other criteria.
HMA currently also expands the window further based on TOT, so the "window" is not static in most ways.
I mostly just want to reduce the number of configurables for v4 too, the trigger alg config is getting obscurely long. For v5 this won't be a problem

aeoranday: Good enough to me! I'll make the changes in here and `triggeralgs`.

ArturSztuc (approved):
LGTM!

PR #439, opened 2024-03-25
New BundleN Configurable

aeoranday:
The new BundleN https://github.com/DUNE-DAQ/triggeralgs/pull/62 TA and TC makers in `triggeralgs` include a new configurable parameter `bundle_number`. This is to keep up with the new configurable.


schema/daqconf/triggergen.jsonnet
```diff
@@ -37,6 +37,7 @@ local cs = {
     s.field("adc_threshold", types.count, default=10000),
     s.field("n_channels_threshold", types.count, default=8),
     s.field("print_tp_info", types.flag, default=false),
+    s.field("bundle_number", types.count, default=100),
```
ArturSztuc: Maybe not for this PR, but perhaps `bundle_size`? `bundle_number` sounds a bit like an ID of a bundle, or even number of bundles.

ArturSztuc (approved):

schema/daqconf/triggergen.jsonnet
aeoranday: That definitely sounds like a change for this PR. Otherwise it may never get changed. I was just extending `BundleN` to `bundle_number`. I had no preference either way.

ArturSztuc: It could always be changed at the merge into develop with the OKS schema implementation, but if it's not too much work (on the background of everything else happening) would you mind changing it?

PR #440, opened 2024-03-25
Adde TPStreamWriter configuration parameters

bieryAtFnal:
The code in this PR adds the following configuration parameters to those supported by `daqconf`:

- tp_accumulation_inactivity_time_before_write_sec, default 1.0 sec
- tp_accumulation_interval_ticks, default 62500000 ticks

The former is branch new, and it can be used to increase the amount of time that the TPStreamWriter waits for all of the TPSets that are stored in a given TimeSlice.

The latter has been around for a while, but only at the `dfmodules` repo level.

I made the new parameter have units of seconds because that seems more easily understandable.

I left the units of the existing pameter in ticks, in the spirit of reducing the number of changes.

This PR is correlated with ones in [dfmodules](https://github.com/DUNE-DAQ/dfmodules/pull/336) and [daqsystemtest](https://github.com/DUNE-DAQ/daqsystemtest/pull/85).


bieryAtFnal:
I believe that Wes has successfully tested these changes at EHN1, and we believe that there is value in going ahead and merging them, so I will do that soon.

PR #441, opened 2024-03-25
CTB: Trigger integration

MRiganSUSX:
Please see https://github.com/DUNE-DAQ/trigger/pull/288 for details.


python/daqconf/apps/trigger_gen.py
```diff
@@ -388,11 +404,17 @@ def get_trigger_app(
...
+        mgraph.connect_modules("ttcm.output",              "tctee_ttcm.input",            "TriggerCandidate", "ttcm_input", size_hint=1000)
+        mgraph.connect_modules("tctee_ttcm.output1",       "mlt.trigger_candidate_input", "TriggerCandidate", "tcs_to_mlt", size_hint=1000)
+        mgraph.connect_modules("tctee_ttcm.output2",       "tc_buf.tc_source",            "TriggerCandidate", "tcs_to_buf", size_hint=1000)
         mgraph.add_endpoint("hsievents", "ttcm.hsi_input", "HSIEvent", Direction.IN)
 
+    if USE_CTB_INPUT:
+        mgraph.connect_modules("ctbtcm.output",              "tctee_ctbtcm.input",          "TriggerCandidate", "ctbtcm_input", size_hint=1000)
+        mgraph.connect_modules("tctee_ctbtcm.output1",       "mlt.trigger_candidate_input", "TriggerCandidate", "tcs_to_mlt",   size_hint=1000)
+        mgraph.connect_modules("tctee_ctbtcm.output2",       "tc_buf.tc_source",            "TriggerCandidate", "tcs_to_buf",   size_hint=1000)
+        mgraph.add_endpoint("hsievents", "ctbtcm.hsi_input", "HSIEvent", Direction.IN)
```
ArturSztuc: I think this will need to be `ctbhsievents`, although the corresponding endpoint in ctb (here: https://github.com/DUNE-DAQ/ctbmodules/blob/f4d5ea7119a17399ff6b9f526309393d00ee891b/python/ctbmodules/apps/ctb_hsi_gen.py#L160) will need to change for that too. @strilov would you like us to change it on the ctb site (do we need to be part of the correct github team?) or someone from CTB will do this?

ArturSztuc (approved):
LGTM in the context of https://github.com/DUNE-DAQ/ctbmodules that exists now, but the endpoint connection will need to change in both repositories for CTB signal so it doesn't get confused with DTS/fake hsi when we're running them simultaneously.

python/daqconf/apps/trigger_gen.py
jsensenig: @ArturSztuc I can make the change in the CTBModule. L160 that you pointed to I can change from `hsievents` to ctbhsievents`.

ArturSztuc: Thank you @jsensenig , looks like @strilov has already made a branch for this here: https://github.com/DUNE-DAQ/ctbmodules/compare/production/v4...feature/individual_hsi_tc_makers
It will probably need to be merged together with this one

jsensenig: Ah I missed that, it looks like it's taken care of then. We'll just need to merge in his changes.

python/daqconf/apps/trigger_gen.py
```diff
@@ -194,7 +196,7 @@ def get_trigger_app(
             TP_SOURCE_IDS[trigger_sid] = conf
        
     # Check for present of TC sources. At least 1 is required
-    if not tc_source_present(USE_HSI_INPUT, USE_CTB_INPUT, USE_CUSTOM_MAKER, USE_RANDOM_MAKER, len(TP_SOURCE_IDS)):
+    if not tc_source_present(USE_HSI_INPUT, USE_FAKE_HSI_INPUT, USE_CTB_INPUT, USE_CUSTOM_MAKER, USE_RANDOM_MAKER, len(TP_SOURCE_IDS)):
```
ArturSztuc, line 199: The `tc_source_present` function itself was not changed, will this line not cause a crash?

python/daqconf/apps/trigger_gen.py
```diff
@@ -340,7 +345,22 @@ def get_trigger_app(
...
+                                                       time_before=TRIGGER_WINDOW_BEFORE_TICKS,
+                                                       time_after=TRIGGER_WINDOW_AFTER_TICKS),
+                                         s1=ttcm.map_t(signal_type=TTCM_S1,
+                                                       time_before=TRIGGER_WINDOW_BEFORE_TICKS,
+                                                       time_after=TRIGGER_WINDOW_AFTER_TICKS),
+                                         s2=ttcm.map_t(signal_type=TTCM_S2,
+                                                       time_before=TRIGGER_WINDOW_BEFORE_TICKS,
+                                                       time_after=TRIGGER_WINDOW_AFTER_TICKS),
+                                         hsi_trigger_type_passthrough=HSI_TRIGGER_TYPE_PASSTHROUGH,
+                                         prescale=TTCM_PRESCALE))]
```
ArturSztuc, line 362: Not for this PR, but I assume we might need separate config inputs for `hsi` vs `fake_hsi`... we'll deal with this once it's needed tho :/

ArturSztuc (commented):
LGTM, apart from the ` tc_source_present` function that needs updating to take the new input

MRiganSUSX:
Here are examples from a test configuration including changes on the hsi / ctb end (thanks @ArturSztuc):
system
system_no
trigger

PR #442, opened 2024-03-26
add back in tpg threshold configuration

wesketchum:
Fixes inability to configure TPG threshold.


adam-abed-abud (approved):
Thanks Wes for the quick PR. This looks good to me, I tested it on two different configs and the tpg threshold appears to be properly set.

bieryAtFnal (approved):
LGTM.

I compared what we see for tpg_threshold values in the 3ru_3df integtest, both before the TPG configurability changes and now (with this change), and they look identical.

PR #443, opened 2024-03-26
update the crt dro to have an input directory option, and other fixes

wesketchum:
This is a draft pull request on adding a 'CRT' stream to the detector readout map, currently needed for trying to run with the CRT. If we find a way to configure the CRT application differently, than this may go away.


alessandrothea (approved):

wesketchum:
All running at EHN1 now, with the MPMC queue in the last commit (which @roland-sipos approved of).

PR #448, opened 2024-04-04
Adding ttcm readout window override

ArturSztuc:
A feature requested by @wesketchum  [HERE](https://dunescience.slack.com/archives/C04BATAPKK7/p1712057016688679):

When user doesn't provide any `TimingTriggerCandidateMaker` configuration, it is still being populated by the default:
```
        {"signal":0, "tc_type_name":"kTiming", "time_before":1000, "time_after":1000},
        {"signal":1, "tc_type_name":"kTiming", "time_before":1000, "time_after":1000},
        {"signal":2, "tc_type_name":"kTiming", "time_before":1000, "time_after":1000},
        {"signal":3, "tc_type_name":"kTiming", "time_before":1000, "time_after":1000}
```

If the user provides the map, but leaves `"time_before"` or `"time_after"` empty for any of the TTCM/HSI signals, they will be set to `-1` by default, and then overridden with `"trigger_window_before_ticks"` and/or `"trigger_window_after_ticks"`. I couldn't figure out an easier way of overriding this in the schema, but if anyone has some ideas, please shout...

So, e.g. user-provided map:
```json
"trigger": {
      "trigger_window_after_ticks": 8000,
      "trigger_window_before_ticks": 2000,
      "ttcm_input_map": [
          {"signal": 0, "time_after": 100, "time_before": 0},
          {"signal": 1, "time_after": 100},
          {"signal": 2},
       ]
}
```

Will result in the actual map of:
```json
{
  "data": {
    "hsi_configs": [
      {"signal": 0, "tc_type_name": "kTiming", "time_after": 100, "time_before": 0},
      {"signal": 1, "tc_type_name": "kTiming", "time_after": 100, "time_before": 2000},
      {"signal": 2, "tc_type_name": "kTiming", "time_after": 8000, "time_before": 2000 }
    ],
    "prescale": 1
  },
  "match": "ttcm_fake"
},
```
The change will only be visible in `data/trigger_conf.json`, not in `config/daqconf.json`, since the override happens after default of `-1` was applied.

Tested by:
1. Everything from https://github.com/DUNE-DAQ/integrationtest/tree/develop/python/integrationtest
2. Successfully generating the few variants of the configs above with `fake_hsi`, running offline and checking with the new `HDF5LIBS_TestDumpRecord` if the TCs have correct window sizes.

### DIFF
--- a/python/daqconf/apps/trigger_gen.py
+++ b/python/daqconf/apps/trigger_gen.py
@@ -108,6 +108,31 @@ def check_mlt_roi_config(mlt_roi_conf, n_groups):
 def tc_source_present(use_hsi, use_fake_hsi, use_ctb, use_ctcm, use_rtcm, n_tp_sources):
 	return (use_hsi or use_fake_hsi or use_ctb or use_ctcm or use_rtcm or n_tp_sources)
 
+
+#===============================================================================
+def update_ttcm_map(ttcm_map,
+                    trigger_window_before_ticks,
+                    trigger_window_after_ticks
+    ):
+    """
+    Populates the readout window for TTCM hsi-TC map with the global values the
+    supplied values are -1 (default if readout window not supplied in TTCM map)
+
+    Args:
+        ttcm_map (dict): The TTCM hsi-TC map as defined in the schema.
+        trigger_window_before_ticks (int): N ticks to expand readout window before event
+        trigger_window_after_ticks (int): N ticks to expand readout window after event
+
+    Returns:
+        dict: Updated TTCM hsi-TC map
+    """
+    for entry in ttcm_map:
+        if entry['time_before'] == -1:
+            entry['time_before'] = trigger_window_before_ticks
+        if entry['time_after'] == -1:
+            entry['time_after'] = trigger_window_after_ticks
+    return ttcm_map
+
 #===============================================================================
 def get_trigger_app(
         trigger,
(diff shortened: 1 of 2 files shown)


bieryAtFnal:
Am I correct in understanding that `trigger_window_before_ticks` and `trigger_window_after_ticks` were deprecated in the first `fddaq-v4.4.0` release candidate, and are now being repurposed (un-deprecated) to provide the defaults for the `ttcm_input_map`?

bieryAtFnal:
I believe that this is working as intended, but I was somewhat confused at first.  

I now realize that it is only explicit `ttcm_input_map` entries in `daqconf.json` that don't have values for `time_before` or `time_after` that get the defaults.  

Initially, I was thinking that one could simply provide values for `trigger_window_before_ticks` and `trigger_window_after_ticks`, and provide *no* config parameters for `ttcm_input_map`, and somehow the values from the former would magically populate the latter.  

The observed behavior is probably fine, but it would be good to ensure that everyone understands how the defaults work.  

I suspect that what confused me was the first part of the description of this PR.  It shows the default values that are used in the code (I believe), and mistook that for values that are defaulted in the generated configuration.

ArturSztuc:
@bieryAtFnal I have modified this now, so the default map uses the global overrides (the un-deprecated `trigger_window_before_ticks`  and `trigger_window_after_ticks`) instead of the individual `time_before` and `time_after` when no config parameters for `ttcm_input_map` is provided.

bieryAtFnal (approved):
These changes look good to me.  With the latest commit, I have looked at the generated configuration, and it looks reasonable.  I also verified that the integtests in daqsystemtest work, both *with* the changes for the ttcm_input_map and *without* them.  (Artur's latest commit opens the door to going back to the state of the daqsystemtest integtests from a couple of weeks ago.  I'm not going to advocate that, but it seemed like a nice way to verify that the defaults are working.)

MRiganSUSX (approved):
This looks good.

(it will conflict with another change in https://github.com/DUNE-DAQ/daqconf/pull/452, but I can update that once this is merged...)

PR #449, opened 2024-04-05
Fake hsi split

MRiganSUSX:
Little config change that allows routing fake hsi signal to either timing or ctb modules.
Notes:
-- fake hsi comes from a direct link (not pubsub) so we can only have them routed to one module
-- this allows testing of other plugins (in addition to ttcm one can also test ctbtcm, in future maybe cibtcm)
-- it is possible to have DTS, CTB and RANDOM (fake) at the same time

Associated PRs: 
- fddaqconf: https://github.com/DUNE-DAQ/fddaqconf/pull/32
- trigger: https://github.com/DUNE-DAQ/trigger/pull/294

Cases:
- Just fake HSI as timing (DTS):

- Just fake HSI as CTB:

- timing + fake CTB:

- timing + fake timing (RANDOM):

- CTB + fake timing:

- all three (timing + CTB + random):

Special cases:
- fake timing and fake ctb: 
This is not possible, the CTB flag is either true or false and therefore just one link will be made

- no fake but flag set for hsi_to_ctb:
Results in error:

### DIFF
--- a/config/daqconf_full_config.json
+++ b/config/daqconf_full_config.json
@@ -97,7 +97,8 @@
         "hsi_source": 1,
         "mean_hsi_signal_multiplicity": 1,
         "use_fake_hsi": true,
-        "use_timing_hsi": false
+        "use_timing_hsi": false,
+        "fake_hsi_to_ctb": false
     },
     "readout": {
         "data_files": [],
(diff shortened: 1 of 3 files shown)


bieryAtFnal:
I haven't tried out the various combinations of modules that Michal describes in the overview of these changes, but I have successfully run all of the integtests from the daqsystemtest and dfmodules repos, and I wanted to report on that success.

bieryAtFnal (approved):
LGTM.
I have not tried out any of the nicely documented scenarios that Michal has described, but I've confirmed that the changes do not affect the successful running of the daqsystemtest integtests, as expected.

ArturSztuc (approved):
LGTM, I tested it and it works! Thank you @MRiganSUSX

PR #450, opened 2024-04-10
Adding interface for the CIB

nbarros:
This PR adds a new HSI-like object as a producer of TCs. The code is largely based on the CTB code, as the two pieces of hardware are very similar. 

By default the CIB is off, so these changes should not affect the operation without the CIB.

edit: tested this change (along with the remaining CIB related code) with `minimal_system_quick_test.py` and no issues were reported.


bieryAtFnal (approved):
LGTM in so far as existing integtests continue to run successfully.

PR #451, opened 2024-04-10
Trigger config cleanup

MRiganSUSX:
removing obsolete config parameters from trigger config


ArturSztuc (approved):
Thank you for this, LGTM!

PR #455, opened 2024-05-07
Threshold By Plane Configurables

aeoranday:
Configurables PR for TPG threshold by plane implementation in [fdreadoutlibs](https://github.com/DUNE-DAQ/fdreadoutlibs/pull/177).

Should `configs/daqconf_full_config.json` and `configs/daqconf_with_tpg.json` also reflect these changes? Leaving `<plane>_threshold` out of a `daqconf.json` will result in that plane defaulting to `tpg_threshold`, so I can see it going with or without.


python/daqconf/apps/readout_gen.py
```diff
@@ -205,6 +205,9 @@ def add_tp_processing(
                         link_id = geo_lid,
                         enable_tpg = True,
                         tpg_threshold = cfg.tpg_threshold,
```
alessandrothea: Remove?

python/daqconf/apps/readout_gen.py
```diff
@@ -205,6 +205,9 @@ def add_tp_processing(
...
+                        collection_threshold = cfg.collection_threshold,
```
alessandrothea: prefix `tpg_`

alessandrothea (changes requested):

aeoranday:
Latest commit made the word choice changes to match what is used in `detchannelmaps`. Configs and variables now refer to planes 0, 1, and 2 instead of induction 1, induction 2, and collection. Also prefixed with `tpg_`.

alessandrothea (approved):

PR #461, opened 2024-05-23
Latency measurement (for opmon) throughout trigger flow

MRiganSUSX:
Associated trigger PR: https://github.com/DUNE-DAQ/trigger/pull/303


ArturSztuc (approved):
Probably should have done `git rebase` after the last release, it's very difficult to tell which changes are yours and which are already in since 2 weeks ago.

I don't like that the offset needs to be placed into each module "manually" in `trigger_gen`, but I don't have any better ideas so LGTM!

PR #462, opened 2024-06-03
Removing TP buffers from the trigger

ArturSztuc:
I removed the `TPbuffer`s from the trigger after a final consensus from the trigger team.
The readout system already buffers TPs, duplicated code is not good for bug-hunting, duplicated buffers cause confusion offline when looking at the saved data, and needlessly create an extra latency for the trigger, <paste your own rant here>. 

Depending on the configuration, for the single `TriggerActivityMaker`/APA it will pass TPs from `TPChannelFilter` straight into `TriggerActivityMaker`, and in case of multiple TAMs/APA, it will send TPs from `TPChannelFilter` to `TPSetTee`, which will split it between the individual TAMs (pasted graph below to show this).
trigger

Tested this by successfully running with the replay application, and by running live on NP04 (with long runs over the weekend, over the nights etc).

Another PR in the trigger that removes the now-unused C++ code: https://github.com/DUNE-DAQ/trigger/pull/306


ArturSztuc:
Closing, we have merged this with Alessandro's branch, PR: https://github.com/DUNE-DAQ/daqconf/pull/463

PR #463, opened 2024-06-06
Mrigan/per plane buffs removed

ArturSztuc:
Alessandro's changes in the readout and trigger applications to enable TP Processing (by the TP handler) by-plane in the trigger, and by-plane TAMaking in the trigger.

We removed the TP buffers from the trigger too, since they are already in the readout. The dedicated PR for that (https://github.com/DUNE-DAQ/daqconf/pull/462) is removed.

This PR goes together with https://github.com/DUNE-DAQ/fdreadoutlibs/pull/185 and https://github.com/DUNE-DAQ/trigger/pull/306

Tested by running successfully on NP04 & reviewing configuration graphs:
trigger


wesketchum (approved):
tested at NP04, and passing integ tests.

MRiganSUSX (approved):
unsure about the .github/workflow but the actual code LGTM.

PR #464, opened 2024-06-06
importing oksconfgen scripts

alessandrothea:
And removing the old daqconf code


python/oksconfgen/assets.py
gcrone, line 1: Shouldn't the `python/oksconfgen`  directory be renamed to `python/daqconf`?

alessandrothea: You're right. Fixed

gcrone (approved):
Looks OK

PR #467, opened 2024-06-11
Renamed dal repos after the latest renaming campaign

alessandrothea:
`appdal`->`appmodel`
`coredal`->`confmodel`


gcrone (approved):
All good

PR #469, opened 2024-06-11
Adding possibility to disable software triggering completely

ArturSztuc:
Added the possibility to disable software triggering completely at the config generation time, with a new option in the trigger configuration: `use_software_triger`, set to true by default (so the default behaviour is unchanged).

This means we can now run TPG, but without all the `TPChannelFilter`s, `TriggerActivityMaker`s, `TriggerCandidateMaker`s, TA buffers, zipper etc. being made. All the custom/external trigger makers (e.g. `TimingTriggerCandidateMaker`, `CIB`/`CTB`etc. trigger candidate makers) are still possible with `use_software_trigger` off.

Tested by:
1. Running systemtest: `3ru_3df_multirun_test.py`
2. Generated trigger graphs with `use_software_trigger` on and off, both with TPG on (attached below)
3. Ran both from 2. offline on one of the asset files, with output as expected: seen software triggers with flag on, and no software triggers (no TAs, no TPChannelFilters etc.) with flag set to off in grafana.

NP04 config with TPG ON and `use_software_trigger` ON:
trigger

NP04 config with TPG ON and `use_software_trigger` OFF:
trigger

### DIFF
--- a/config/daqconf_full_config.json
+++ b/config/daqconf_full_config.json
@@ -137,6 +137,7 @@
         "timing_session_name": ""
     },
     "trigger": {
+        "use_software_trigger": true,
         "host_trigger": "localhost",
         "mlt_buffer_timeout": 100,
         "mlt_ignore_tc": [],
--- a/python/daqconf/apps/trigger_gen.py
+++ b/python/daqconf/apps/trigger_gen.py
@@ -170,6 +170,7 @@ def get_trigger_app(
                                      trigger.trigger_window_before_ticks,
                                      trigger.trigger_window_after_ticks)
     TTCM_PRESCALE=trigger.ttcm_prescale
+    USE_SOFTWARE_TRIGGER = trigger.use_software_trigger
     USE_HSI_INPUT = use_hsi_input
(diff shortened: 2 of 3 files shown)


ArturSztuc:
@wesketchum & @bieryAtFnal   I cannot add this to the tracking projects: https://github.com/orgs/DUNE-DAQ/projects/5. 
Getting the following error:
```
Projects cannot have more than 1200 items. To add more, please archive or delete existing items.
```

bieryAtFnal:
@ArturSztuc thank you very much for the heads-up.
In the Tracking Project, I've archived a good number of completed PRs, so we should now be able to assign PRs and Issues to it again.  
I will update the fields in this PR.

MRiganSUSX:
I have tested generating configurations with all the conditions ("enable_tpg": true/false, "use_software_triger": true/false) and the configurations all generated without errors and the .dot for trigger are all correct. 

The only suggestion is to remove the print (or only print it out when debug is on).

MRiganSUSX (approved):
lgtm

PR #473, opened 2024-06-13
Unique and short thread names for trigger's DAQModules

ArturSztuc:
Renaming trigger module's unique names to be less than 16 chars, as needed for threads names (explained [HERE](https://linux.die.net/man/3/pthread_setname_np)).

PR fully explained in the connected trigger PR: https://github.com/DUNE-DAQ/trigger/pull/310

### DIFF
--- a/python/daqconf/apps/trigger_gen.py
+++ b/python/daqconf/apps/trigger_gen.py
@@ -538,9 +538,9 @@ def get_trigger_app(
                 plane = f'{tp_conf.plane}'
               
                 if USE_CHANNEL_FILTER:
-                    mgraph.add_endpoint(f"tpsets_{ru_sid}", f"tpcf_{region}_{plane}.tpset_source", "TPSet", Direction.IN, is_pubsub=True)
+                    mgraph.add_endpoint(f"tpsets_tplink{ru_sid}", f"tpcf_{region}_{plane}.tpset_source", "TPSet", Direction.IN, is_pubsub=True)
                 else:
-                    mgraph.add_endpoint(f"tpsets_{ru_sid}", f'tam_{region}_{plane}_0.input', "TPSet", Direction.IN, is_pubsub=True)
+                    mgraph.add_endpoint(f"tpsets_tplink{ru_sid}", f'tam_{region}_{plane}_0.input', "TPSet", Direction.IN, is_pubsub=True)
 
         for (region_id, plane), ta_conf in TA_SOURCE_IDS.items():
             buf_name = f'ta_buf_{region_id}_{plane}'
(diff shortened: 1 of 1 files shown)


python/daqconf/apps/trigger_gen.py
```diff
@@ -245,16 +245,16 @@ def get_trigger_app(
                           plugin = 'TCBuffer',
                           conf = get_buffer_conf(TC_SOURCE_ID["source_id"], DATA_REQUEST_TIMEOUT))]
     if USE_HSI_INPUT:
-        modules += [DAQModule(name = 'tctee_ttcm',
+        modules += [DAQModule(name = 'tctee_t',
```
MRiganSUSX, line 248: This naming seems inconsistent (fake hsi = tcmfake but real hsi (timing) = t?); but it is propagated correctly (just checking you wanted the names like this).

python/daqconf/apps/trigger_gen.py
```diff
@@ -282,14 +282,16 @@ def get_trigger_app(
...
-        for tp_sid in TP_SOURCE_IDS:
-            link_id = f'tplink{tp_sid}'
+        for tp_sid,tp_conf in TP_SOURCE_IDS.items():
+            ru_sid = f'{tp_conf.tp_ru_sid}'
+            region = f'{tp_conf.region_id}'
+            plane = f'{tp_conf.plane}'
 
             if USE_CHANNEL_FILTER:
-                modules += [DAQModule(name = f'channelfilter_{link_id}',
+                modules += [DAQModule(name = f'tpcf_{region}_{plane}',
```
MRiganSUSX, line 294: Not a code issue, just an idea that we should perhaps make a legend somewhere (github doc?) for the different acronyms for our modules / thread names (specifically tpcf - while somewhat obvious to us - may not be immediately recognised).

python/daqconf/apps/trigger_gen.py
```diff
@@ -339,11 +341,11 @@ def get_trigger_app(
                                                           window_time=10000,  # should match whatever makes TPSets, in principle
                                                           buffer_time=10*ticks_per_wall_clock_s//1000, # 10 wall-clock ms
                                                           activity_maker_config=temptypes.ActivityConf(ACTIVITY_CONFIG[j]))),
-                                DAQModule(name = f'tasettee_region_{region_id}_{plane}_{j}', plugin = "TASetTee")]
+                                DAQModule(name = f'tasettee_{region_id}_{plane}_{j}', plugin = "TASetTee")]
```
MRiganSUSX, line 344: This could be part of the same legend mentioned above, that the naming follows `region_plane_id`, since we are removing the region from the string name (which I agree is a good thing).

MRiganSUSX (approved):
No code issues. 
Some comments to check whether you wanted it that way.
Some notes on possible future additions. 

Functionally can be used. Tested offline and with integtests (minimal, fake producer, 3ru 1df)

ArturSztuc:
I think a "legend" is a good idea. At the very least, we should update trigger graphs in the documentation we have

PR #490, opened 2024-07-22
Asztuc/ta to tc workflow

ArturSztuc:
Changing the `TAM->TASet->TCM` `TriggerActivity` workflow to `TAM->TA->TCM` workflow, removing `TASets`, having individual TAs being sent as soon as they are made, and removing the `TAZipper`.
Full description in https://github.com/DUNE-DAQ/trigger/pull/327


MRiganSUSX (approved):
LGTM!

PR #493, opened 2024-07-24
Allowing ADCSimpleWindow user-provided tc-type output

ArturSztuc:
This PR allows the ADCSimpleWindow to produce TC with any user-provided TC-type (something that in v5 we will probably want all algorithms to be capable of), and fixes a small bug for creating configuration with multiple triggering algorithms per subdetector. Full description in https://github.com/DUNE-DAQ/triggeralgs/pull/75


aeoranday (approved):
LGTM.

PR #496, opened 2024-09-04
Add simple script to reformat OKS datbases to ensure that objects are…

eflumerf:
… in OKS's preferred order


henry-wallace-phys (approved):
Yep still looks fine

PR #497, opened 2024-09-10
add utility to enable/disable TPG

glehmannmiotto:


eflumerf (approved):
Works with test-session

PR #501, opened 2024-09-18
Adapt to schema changes for readout app tp_source_ids and sessions op…

gcrone:
…mon uri


eflumerf (approved):

eflumerf:
Only potential alteration is trying to get an opmon_uri from the config before making one.

PR #503, opened 2024-09-18
Instead of assuming that all VirtualHost entries in the configuration

eflumerf:
are available for running readout applications, take an array of allowed hosts as a parameter, putting all readout apps on localhost if none are specified.


gcrone (approved):
LGTM

PR #504, opened 2024-09-19
Removed the tc-maker-1 app from generate_trigger, for now…

bieryAtFnal:
…since it does not seem to be needed, and it causes spurious warning messages in the minimal_system_quick_test.py log files.   

I've made tentative changes in this repo and in the daqsystemtest repo (same branch name).  I've confirmed that the minimal_system_quick_test.py continues to work with these changes.

The changes that I've made here may be too brute-force.  I can imagine that we might want an option to the generate_trigger(OKS) function that specifies whether the tc-maker-1 app should be included or now.  Also, it would be great for someone to confirm that I've sliced out the right amount of tc-maker-1-related lines (not too much or too little).

PR #505, opened 2024-09-19
Use tpg_enabled to determine whether to generate tc-maker-1 app in ge…

### LINKED ISSUE #114: Why does running the minimal_system_quick_test modify config/daqsystemtest/ccm.data.xml?
It seems like it would be nice if integtests do not modify code in locally-cloned repos...

eflumerf:
…nerate_trigger. Add copy_configuration utility to copy all included databases to a directory, updating includes as needed.


eflumerf:
Replaces #504

PR #506, opened 2024-09-25
Candidate script to set the value of a DAQ session env var in a specified configuration

bieryAtFnal:


glehmannmiotto:
I think that drunc is being modified to set those variables automatically for the apps, so the values don't need to be in the OKS DB.

alessandrothea:
I believe that logic is being put in place in drunc to capture (some) environment variables but I don't think that automatically implies in the removal of environment variables from the configuration. If one takes nanorc as an example, relevant environment variables were defined in the configuration (`boot.json`). For each variable `boot.json` specified either a value or that the variable should be taken from the environment. Unless there is a different proposal, I think a similar approach would be quite effective.

alessandrothea:
@bieryAtFnal a small request: could you rename the script `daqconf_set...`? Thank you.

plasorak:
> I believe that logic is being put in place in drunc to capture (some) environment variables but I don't think that automatically implies in the removal of environment variables from the configuration. If one takes nanorc as an example, relevant environment variables were defined in the configuration (boot.json). For each variable boot.json specified either a value or that the variable should be taken from the environment. Unless there is a different proposal, I think a similar approach would be quite effective.

That's correct, some variables will still be in the configuration. There is right now no possibility to pass your own env to the processes, that can be added if needed (although we'd need to decide which ones and how - like in nanorc `getenv/getenv_ifset` I suppose?).

There is however in the possibility to pass, in boot, some override env variables:
```
boot -e FOO=BAR -e BAZ=BAR
```

Let me write reopen my PR in drunc, and I'll write up what I've done there.

PR #520, opened 2024-10-11
Asztuc/triggerbuffer resize

ArturSztuc:
Resizing trigger buffers to be more reasonable and not take 100s of GBs of RAM.


MRiganSUSX (approved):
I can confirm this successfully changes the `latency_buffer_size` in `latencybufferconf` and `requesthandlerconf` for all TA buffers and the TC buffer and it does not change anything else.

ran over the usual set of integration tests.

eflumerf:
This appears to be obsolete. Is there any plan for a v4.4.x patch release in the future? If not, we should probably close. @ArturSztuc @MRiganSUSX

ArturSztuc:
On one hand this is an important fix for v4, on the other hand I would be upset if we ended up using v4 anyway.
I'll close it, and if @wesketchum or someone else says that we will run v4 on anything then we can re-open.

PR #522, opened 2024-10-14
Add option to turn off Connectivity Service as an infrastructure app

eflumerf:
Also, support disabling the connectivity service completely


bieryAtFnal (approved):
I've verified that a similar fraction of the automated regression tests in the daqsystemtest repo pass, both with and without this change (and the [companion change in the integrationtest repo](https://github.com/DUNE-DAQ/integrationtest/pull/84)).  I believe that the failures in the integtests that I observed were unrelated to this change.

PR #523, opened 2024-10-17
Added logic in generate.py to add the value of the TRACE_FILE env var…

bieryAtFnal:
… to the OKS Session, if that env var is set in the user's shell environment.

This will allow us to have control (e.g. enable and view) TRACE debug messages in our integtests.  

My sense is that this will *not* affect interactive running.  I believe that for interactive running, one will still need to add the TRACE_FILE Variable to the OKS Session (soon to be OKS System) in order to enable TRACE debug messages within a given system configuration.

I tested this change by verifying that TRACE messages appeared in the TRACE memory buffer when the minimal_system_quick_test was run with the TRACE_FILE env var set in the Linux shell environment.

### DIFF
--- a/python/daqconf/generate.py
+++ b/python/daqconf/generate.py
@@ -1128,6 +1128,12 @@ def generate_session(
 
     opmon_svc = db.get_dal(class_name="OpMonURI", uid="local-opmon-uri")
 
+    trace_file_var = None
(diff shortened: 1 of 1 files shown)


eflumerf (approved):
🙉

PR #530, opened 2024-10-28
Added a script to allow us to set the ConnectivityService port to a random available port number

bieryAtFnal:
I also cleaned up some unneeded imports in the set_session_env_var script(s).

So far, I've only tested this on the local-1x1-config OKS Session in the daqsystemtest/config/daqsystemtest/example-configs.data.xml database.  I'll do more, and testing from others would be appreciated.


python/daqconf/randomize_connectivity_service_port.py
```diff
@@ -0,0 +1,57 @@
...
+    db.update_dal(session.connectivity_service.service)
+
+    for app in session.infrastructure_applications:
+        #print(f"{app}")
+        if app.className() == "ConnectionService":
+            index = 0
+            for clparam in app.commandline_parameters:
+                if "gunicorn" in clparam:
+                    pattern = re.compile('(.*0\.0\.0\.0\:)\d+(.*)')
+                    blah = pattern.sub('\\1PORT\\2', clparam)
```
eflumerf: Can these three lines be `app.commandline_parameters[index] = pattern.sub(f'\\1{new_port}\\2',clparam)`

eflumerf: It's also possible that the `(.*)` at the end is not needed

eflumerf (approved):
Works for me, started calling the function from `integrationtest`.

PR #535, opened 2024-11-05
JCF: Issue #534: deal with loopbacks in application plots

### LINKED ISSUE #534: create_config_plot does not correctly deal with subscribers with both internal and external publishers
Specifically noticed with the `tcs_.*` subscription in the MLT app. When plotting the whole example-configs/local-1x1-config configuration, `tcs_.*` is correctly shown going both from MLT to MLT and from hsi-to-tc-app to MLT. When plotting just the mlt, however (-r mlt), only the internal instance of `tcs_.*` is shown (the external inhttps://files.slack.com/files-pri/T03RN7KU3-F07V0PQSDNY/screenshot_2024-11-05_at_9.20.43___am.pngput from hsi-to-tc-app is missing).
Screenshot 2024-11-05 at 9 21 25 AM
Screenshot 2024-11-05 at 9 20 43 AM

jcfreeman2:
The problem this commit deals with is basically, if you have a module which subscribes on the network to a wildcarded message (say "msg_.*") and another module in the same application which sends a version of message that matches the wildcard (say "msg_A") out onto the network, then how do you plot that?

Previously, when plotting, the logic in the plotter didn't care about whether a message was sent onto the network. So it was drawing a connection between the modules making it look as if they communicated via a queue. With this commit, if we're plotting an application, a message gets correctly represented as being sent out over the network, and as being received from the network (with the usual "X" and "O" to represent "region outside of this diagram").


eflumerf (approved):
I think this is an improvement. Now the potential issue is that users must connect the outgoing connection to the incoming one themselves

eflumerf:
Screenshot 2024-11-06 100937

bieryAtFnal (approved):
In the new version of the MLT-focussed view, at first it wasn't obvious to me what I was seeing.  However, after re-rereading Eric's comments and looking at it a little more, I figured it out.
In the end, I agree that this is an improvement.  It will probably be good to educate people that sometimes we need to look for fully-specified outputs (e.g. X = tcs_random-tc-generator) that match wildcard inputs (e.g. O = tcs_*) to get all of the connections clear...

PR #537, opened 2024-11-08
Textual Based Database Editor

henry-wallace-phys:
Full implementation of a textual python database editor. Has ability to modify object attributes/relationships etc.


eflumerf:
I think this is more suited for `develop`

PR #539, opened 2024-11-11
Update `consolidate` doc

plasorak:
... To be correct, the `consolidate` utility does not inlcude schema files.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -18,7 +18,7 @@ Commandline utility to generate a graphical flow diagram of a full configuration
   Add Resource objects to or remove from the `disabled` relationship of a Session
 
 ### `consolidate`
-  Merge the contents of several database files, putting all objects into a single output file. Output file will only include schemas.
+  Merge the contents of several database files, putting all objects into a single output file. Output file will NOT include schemas.
 
 ### `consolidate_files`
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -18,7 +18,7 @@ Commandline utility to generate a graphical flow diagram of a full configuration
   Add Resource objects to or remove from the `disabled` relationship of a Session
 
 ### `consolidate`
-  Merge the contents of several database files, putting all objects into a single output file. Output file will only include schemas.
+  Merge the contents of several database files, putting all objects into a single output file. Output file will NOT include schemas.
```
eflumerf: I think a better clarification would be "The output file's include list will contain the schema files included by the source databases (or their includes), and will not contain any object databases."

plasorak: I've just updated that comment.

eflumerf: I still think there's too much confusion between the "object schema includes" and "object database includes".
consolidate will have only object schema includes, i.e. OKS-formatted xml files with extension .schema.xml
consolidate_files will have both object schema includes *and* object database includes (.schema.xml and .data.xml). The include list will be equivalent to the _set_ of includes from the input files, minus any input files that were themselves included by other input files.

PR #542, opened 2024-11-14
Cider-Patches

henry-wallace-phys:
Small patch to Cider's disable object interface. Incorrectly displayed that objects that can be disabled couldn't and interface was positioned incorrectly


eflumerf:
I'm assuming this follows/replaces #541? Is there an equivalent PR onto `develop`?

alessandrothea:
This branch is based on `develop`, so not suitable for merge to patch v5.2.1

PR #544, opened 2024-11-19
Unique control service

plasorak:
This PR makes use of https://github.com/DUNE-DAQ/daqsystemtest/pull/162 and https://github.com/DUNE-DAQ/confmodel/pull/53 to simplify the `generate` module (which used by the integration tests).

Note that this PR is against `develop` but the `confmodel` and `daqsystemtest` PRs are against their respective `patch/fddaq-v5.2.x` branches. To get this to work, I merged the branches of `confmodel` and `daqsystemtest` into `develop` of their repo. In case we want this to go to the patch branch, we'll have to cherry-pick.

### DIFF
--- a/python/daqconf/generate.py
+++ b/python/daqconf/generate.py
@@ -54,8 +54,7 @@ def generate_dataflow(
         hosts.append("vlocalhost")
 
     # Services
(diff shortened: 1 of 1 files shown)


eflumerf (approved):
Had to check out `appmodel` for testing. Works after merging develop

PR #550, opened 2024-11-25
Remove topological sort

henry-wallace-phys:
Removes unnecessary topological sort from CIDER, this was a vestigial feature from the original UI plan.

### DIFF
--- a/python/daqconf/cider/app_structures/main_screen.py
+++ b/python/daqconf/cider/app_structures/main_screen.py
@@ -26,12 +26,12 @@ class MainScreen(Screen):
     BINDINGS = [
                 # Binding("ctrl+s", "save_configuration", "Save Configuration"),
                 Binding("ctrl+s", "save_configuration_with_message", "Save Configuration"),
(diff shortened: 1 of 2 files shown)


python/daqconf/cider/app_structures/main_screen.py
```diff
@@ -26,12 +26,12 @@ class MainScreen(Screen):
     BINDINGS = [
                 # Binding("ctrl+s", "save_configuration", "Save Configuration"),
                 Binding("ctrl+s", "save_configuration_with_message", "Save Configuration"),
```
alessandrothea, line 28: Any reason to keep the Ctrl modifier for save and quit?

henry-wallace-phys: For save it's just because ctrl+s is the standard save option. For quit+delete there's no reason beyond it being slightly harder to accidentally hit those keys and bring up an annoying popup menu

alessandrothea (approved):

PR #552, opened 2024-12-05
Fix some dodgy disable object logic

henry-wallace-phys:
Fix disabled object logic error


alessandrothea (approved):
LGTM

eflumerf:
This was never merged into the patch branch. Was a similar change made to `develop`? Should this PR be re-targeted?

PR #562, opened 2025-03-15
Re-worked the set_session_env_var script to handle the case in which …

bieryAtFnal:
…the requested env var already exists in the specified OKS Session.

Previously, the script would just blindly create a new OKS Variable and add it to the OKS Session, even if one already existed for the requested env var.  Obviously, this was not good.

In the changes included here, if a new OKS Variable needs to be created, it is given a name of the form "temporary-env-<requested_env_var_name>".  The "temporary" prefix is intended to convey that it was created by a script and should not be committed as an official change without additional consideration by a developer (and likely a renaming).

### DIFF
--- a/python/daqconf/set_session_env_var.py
+++ b/python/daqconf/set_session_env_var.py
@@ -1,7 +1,7 @@
 import conffwk
 import confmodel
 
-def set_session_env_var(oksfile, env_var_name, env_var_value, session_name):
+def set_session_env_var(oksfile, session_name, requested_env_var_name, requested_env_var_value):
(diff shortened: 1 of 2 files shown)


MRiganSUSX (approved):
Thanks for this change Kurt. 

I can confirm that, using the script:
- if the variable exists, its value is successfully changed as requested
- if the variable does not exist, a new (temporary) one is created

I've also run few integtests, not that this change should affect them in any way.

Perhaps for a future change/PR: it would be nice to see some kind of confirmation/feedback when the script is run (whether it found and changed the var, whether it created new one, how the value has changed...).

PR #566, opened 2025-04-30
Modified the generate_readout function inside the generate.py file to…

bieryAtFnal:
… only create one TP data source for non-TPC detectors.

This change came about as part of trying to restore the PDS data-type test in the readout_type_scan integtest in daqsystemtest and add PDS TPs to that integtest.

When I ran some tests before I made this change, three TP-related modules were created in each ReadoutApp, but from what I see in configurations at NP02, only one TP-related module should be created for PDS/DAPHNE TPs.  And, with this change, I see the correct behavior in the PDS_TPG part of readout_type_scan.

### DIFF
--- a/python/daqconf/generate.py
+++ b/python/daqconf/generate.py
@@ -648,7 +648,9 @@ def generate_readout(
             ru.tp_handler = tphandler
             tp_sources = []
             tpbaseid = (appnum * 3) + 100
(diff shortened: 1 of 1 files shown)


eflumerf (approved):
LGTM. The "magic numbers" here really should be replaced at some point.

PR #568, opened 2025-05-15
JCF: limit the set of relations between objects which qualify as pare…

jcfreeman2:
…nt-child relations


jcfreeman2:
The `GraphBuilder::find_child_objects` function is used to establish parent-child relations between sessions, segments and applications. Without the code on this PR, _any_ relation between these object type (e.g., the `disabled` relation) is considered to establish this, which is incorrect.

mroda88 (approved):
The changes allowed the plotting of all the EHN1 configurations to work correctly. Here is an example of one. 

np02-session-np02-session (1)

I admit I haven't looked at all the plots. But all the generations completed successfully.

PR #569, opened 2025-05-15
Johnfreeman/refine parent child relations

jcfreeman2:


jcfreeman2:
The `GraphBuilder::find_child_objects` function is used to establish parent-child relations between sessions, segments and applications. Without the code on this PR, _any_ relation between these object type (e.g., the `disabled` relation) is considered to establish this, which is incorrect.

jcfreeman2:
Closing this right after opening it; it was ill-conceived as it would bring in other changes on the v5.3.x patch branch.

PR #574, opened 2025-05-20
Restored a change in generate.py that somehow was lost in a recent re…

bieryAtFnal:
…vert or merge.

The first indication that something was wrong was in a failure in the overnight running of the readout_type_scan regression test.

It's not clear to me what caused the relevant section of code to go missing, but this PR brings it back on the develop branch.

### DIFF
--- a/python/daqconf/generate.py
+++ b/python/daqconf/generate.py
@@ -648,7 +648,9 @@ def generate_readout(
             ru.tp_handler = tphandler
             tp_sources = []
             tpbaseid = (appnum * 3) + 100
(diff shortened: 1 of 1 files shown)


bieryAtFnal:
To test this, one can create a software area based on a recent nightly build (20-May or later) and run the readout_type_scan_test (e.g. `daqsystemtest_integtest_bundle.sh -f1 -l1`).  Without this change, there will be a failure reported in the DAPHNE_TPG test.  With this change, that failure goes away.

mroda88 (approved):

PR #575, opened 2025-05-27
Inspector `show-resources` command

alessandrothea:
Adding new command to show the enable/disable status of resources and contained resources (in the case of sets)


mroda88 (approved):
LGTM and it was tested

PR #582, opened 2025-07-03
Added an option to set_session_env_var.py to allow users to specify that existing env var values should not be over-written

bieryAtFnal:
This change was needed to support a change in the `integrationtest` repo ([PR 112](https://github.com/DUNE-DAQ/integrationtest/pull/112)) in which I added the setting of the TRACE_FILE env var in integtest configurations that do not already have that env var set.  Please see the `integrationtest` PR for more details.

I also modified the DAL name of the temporary Variables that are created in the configuration to hold the TRACE_FILE env var values to include the OKS Session name.  

Previously, a constant string was used, and this meant that more than one Session could share the same value for the TRACE_FILE env var.  It seemed safer to have dedicate DAL Variables for each Session.

### DIFF
--- a/python/daqconf/set_session_env_var.py
+++ b/python/daqconf/set_session_env_var.py
@@ -1,7 +1,7 @@
 import conffwk
 import confmodel
 
-def set_session_env_var(oksfile, session_name, requested_env_var_name, requested_env_var_value):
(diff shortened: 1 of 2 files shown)


eflumerf (approved):
Tested via integrationtest PR

PR #584, opened 2025-07-16
Tweaked validate script to recognise NP02ReadoutApplications as RO apps

alessandrothea:
Validate would not recognise `NP02ReadoutApplications` as `ReadoutApplications` and throw an error.
Now it does.


mroda88 (approved):
It works for me with the newly created configurations for 5.3.3

PR #587, opened 2025-08-13
Create an object of type Service since type Resource doesn't work any

eflumerf:
more (it became abstract)


bieryAtFnal (approved):
I confirm that with this change, the `oks-format` utility works again.

PR #596, opened 2025-09-08
Use oks python bindings to reformat schema files

eflumerf:
## Description

The `oks-format` utility has been available for formatting `*.data.xml` files for quite some time.  It would be great if it could also be used for formatting `*.schema.xml` files.  The changes in this PR add that functionality.

## Testing Suggestions

Without these changes, running `oks-format` on a `*.schema.xml` file produces an error.  With these changes, the schema file is successfully formatted. 

Also, with these changes, running `oks-format` on a file that is not a `*.data.xml` or `*.schema.xml` file will produce a simple message stating that the script doesn't know how to handle the input file, rather than a trace-back.

### DIFF
--- a/python/daqconf/oks_format.py
+++ b/python/daqconf/oks_format.py
@@ -1,12 +1,24 @@
 import conffwk
+import oks
 
 def oks_format(input_file) -> None:
-    print(f"Formatting file {input_file}")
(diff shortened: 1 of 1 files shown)


bieryAtFnal (approved):
LGTM.  I ran the tests listed in the Testing Suggestions, and those worked as expected.

PR #599, opened 2025-10-03
Change confmodel to confmodel_dal in usage not just in import

gcrone:


jcfreeman2 (approved):
These are a set of straightforward improvements to the status quo. Approved. Having said that, there are further places in this repo we need to modernize to use `confmodel_dal` and `appmodel_dal`; I'll put in a PR for those further improvements after merging this.

PR #602, opened 2025-10-08
SmartDaqApplication inheritance

gcrone:
Adapt to new `generate_modules` signature

Part of issue DUNE-DAQ/appmodel#247
Requires PR DUNE-DAQ/appmodel#248


alessandrothea (approved):
LGTM

PR #603, opened 2025-10-09
Updating the set connectivity script to work with ccm config changes

MRiganSUSX:
Updating the set_connectivity_service_port functions to work with the dynamic port changes in oks. 
Related PR: https://github.com/DUNE-DAQ/daqsystemtest/pull/244

### DIFF
--- a/python/daqconf/set_connectivity_service_port.py
+++ b/python/daqconf/set_connectivity_service_port.py
@@ -1,21 +1,6 @@
 import conffwk
 import confmodel_dal
-import socket
-import random
-
-def find_free_k8s_nodeport():
-    """
-    Finds and returns an available TCP port within the k8s NodePort range (30000-32767).
-    """
-    while True:
-        port = random.randint(30000, 32767)
-        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
-            try:
-                s.bind(("0.0.0.0", port))
-                return port
-            except OSError:
-                continue
-
+from daqconf.utils import find_free_port
 
 def set_connectivity_service_port(oksfile, session_name, connsvc_port=0):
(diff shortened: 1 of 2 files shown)


bieryAtFnal:
I find that the lines that were commented out with

```
 # This block has been removed as it was redundant and caused the confusing printout.
```

are needed to get the example_system_test to run successfully...

MRiganSUSX:
Now updated to work with example_system_test.py.

bieryAtFnal:
I verified that a `local-1x1-config` system run with `drunc` interactively now works fine (with a modified connectivity service port).  And, the `example_system_test` works fine.

My sense is that it would be better to enhance the existing `find_free_port` utility to support a required range of ports, though, instead of adding a new function just for K8s.  I can work on that, if it would be helpful...

MRiganSUSX:
@bieryAtFnal Please, go ahead. I can also rework it if you are very time limited, but would need more detailed description of the preferred outcome.

bieryAtFnal:
I have re-merged the `develop` branches in this repo and the `daqsystemtest` repo to the `mrigan/k8s_changes` branches so that they can be tested with the latest nightly build...

In addition, I've created a branch named `kbiery/tweaks_to_k8s_changes` in this repo on which I've made some changes that I'd like to propose that we make to the original changes in this PR.  My changes combine the `find_free_k8s_nodeport` and `find_free_port` functions into a single one that supports an optional range, and they promoted a few warnings to errors in `set_connectivity_service_port.py`.  I've verified that our usual regression tests work with those proposed changes.  @MRiganSUSX , when you get a chance, please let me know what you think.

MRiganSUSX:
Superseded by https://github.com/DUNE-DAQ/daqconf/pull/608.

PR #604, opened 2025-10-09
SNB Transform demonstrator

eflumerf:
Part of: https://github.com/DUNE-DAQ/daq-deliverables/issues/186

Adds SNB Transform awareness to generate_readout and generate_trigger. If the data receiver type is "FileReaderRecevier", then generate_readout will create a SNB Transform configuration. Likewise, if a configuration exists for the PreconfiguredTriggerModule, then generate_trigger will use that module instead of HSI or random triggers.


andrewmogan (approved):
LGTM.

See comments in [daq-deliverables #189](https://github.com/DUNE-DAQ/daq-deliverables/issues/186) for tests performed.

PR #605, opened 2025-10-13
Added the ability to specify crate and slot numbers for generate_hwmap()…

bieryAtFnal:
…This is useful in cases in which the crate and slot of replay data matters, for example, when generating TPs using a real detector channel map.

## Description

These changes are part of DUNE-DAQ/daqsystemtest#243, and they consist of the addition of the ability to specify a crate_id and slot_id to the `generate_hwmap` code.  This is needed for cases in which we want to have the GeoId (e.g. crate_id and slot_id) in fake-data-replay match what was assigned to the data when it was acquired from the real electronics.

## Testing Suggestions

Please see the instructions listed in DUNE-DAQ/daqsystemtest#243.

### DIFF
--- a/python/daqconf/generate_hwmap.py
+++ b/python/daqconf/generate_hwmap.py
@@ -6,7 +6,7 @@ import json
 import sys
 
 def generate_hwmap(oksfile, n_streams, n_apps = 1, det_id = 3, app_host = "localhost",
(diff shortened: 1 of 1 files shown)


eflumerf (approved):
daqsystemtest_integtest_bundle.sh performed normally

PR #606, opened 2025-10-17
Modified the default frame file provided in generate_readout…

bieryAtFnal:
…to reference a more recent binary file.

## Description

DUNE-DAQ/daqsystemtest#245 requests that we update our regression tests, etc. to use recent data-replay files from the _assets_ system.

In this repo, that translated to changing the checksum of the default asset file that is used in integration tests.  The new file is `wib_link_67.bin`, which was recently created by Eric.

*** Please Note that the change that was made in this repo was modified based on a suggestion from Eric.  Instead of updating the default frame_file value, we removed the default value altogether since it effectively is never used.

## Testing Suggestions

There are suggestions in DUNE-DAQ/daqsystemtest#245 for testing all of the asset-file-update changes together.

### DIFF
--- a/python/daqconf/generate.py
+++ b/python/daqconf/generate.py
@@ -359,7 +359,7 @@ def generate_readout(
     oksfile,
     include,
     generate_segment,
-    emulated_file_name="asset://?checksum=e96fd6efd3f98a9a3bfaba32975b476e",
+    emulated_file_name="asset://?checksum=370df564205290d27cab47e44ae4ca47",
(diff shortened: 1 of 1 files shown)


eflumerf (approved):
Update looks good. I'm wondering if we should define a default value for this parameter here, as the default is already defined in integrationtest/data_classes

bieryAtFnal:
As Eric suggested, I removed the unneeded default value for the emulated data file name.

